PR #6, opened 2024-01-31
GH200 systems require open GPU kernel module driver

mikemckiernan:
- Clarify the RN that GH200 requires the NVIDIA Open GPU Kernel module driver.
- Update the supported GPUs table with the same requirement information.


mikemckiernan:
Francis approved in Slack

PR #7, opened 2024-02-01
Remove blossom-ci.yaml

elezar:


ArangoGutierrez (approved):

elezar:
@ArangoGutierrez @mikemckiernan the actions around docker files need to be reworked too. Since we're just removing the blossom-ci in this PR the failures are definitely unrelated.

One thing I noticed is that the `dockerfile-changed` job should probably just use the docker login action instead of logging in manually.
```
      - name: Log in to Container Registry
        uses: docker/login-action@v2
        with:
          registry: ${{ env.REGISTRY }}
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}
```

PR #33, opened 2024-03-20
Fix install-gpu-ocp.rst when disabling monitoring

chiragkyal:
To disable monitoring, the command should use `openshift.io/cluster-monitoring=false` instead of `openshift.io/cluster-monitoring=true`

### DIFF
--- a/openshift/install-gpu-ocp.rst
+++ b/openshift/install-gpu-ocp.rst
@@ -366,7 +366,7 @@ Disable cluster monitoring in the ``nvidia-gpu-operator`` namespace by setting `
 
    .. code-block:: console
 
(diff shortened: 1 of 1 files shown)


mikemckiernan:
@chiragkyal , do you mind amending your commit with your sign-off and repushing?  Thanks very much for the fix!

chiragkyal:
@mikemckiernan  Done. I've amended sign-off. Thanks!

ArangoGutierrez (approved):

mikemckiernan:
@chiragkyal, thank you again for the fix!

https://docs.nvidia.com/datacenter/cloud-native/openshift/latest/install-gpu-ocp.html#cluster-monitoring

PR #34, opened 2024-03-25
Support 1.28 on EKS

mikemckiernan:
Kubernetes versions: https://nvidia.github.io/cloud-native-docs/review/pr-34/gpu-operator/latest/platform-support.html#supported-operating-systems-and-kubernetes-platforms

### DIFF
--- a/gpu-operator/platform-support.rst
+++ b/gpu-operator/platform-support.rst
@@ -351,12 +351,12 @@ The GPU Operator has been validated in the following scenarios:
            | Kubernetes Service
 
        * - Ubuntu 20.04 LTS
-         - 1.25, 1.26
+         - 1.25---1.28
          - 1.24, 1.25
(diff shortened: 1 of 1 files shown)


gpu-operator/platform-support.rst
```diff
@@ -351,12 +351,12 @@ The GPU Operator has been validated in the following scenarios:
            | Kubernetes Service
 
        * - Ubuntu 20.04 LTS
-         - 1.25, 1.26
+         - 1.25---1.28
          - 1.24, 1.25
```
elezar, line 355: Question: Does it make sense to also use `1.24---1.25` here?

mikemckiernan: I feel like the comma is better when there are only two values.  Though, I concede that it would not be any less clear.

PR #36, opened 2024-04-05
Show Toolkit Customization Syntax in GPU Operator installation

geekloper:
This PR aims to enhance the documentation by adding instructions for passing options to the GPU Operator during installation. 
Please review the changes and provide feedback.

### DIFF
--- a/gpu-operator/getting-started.rst
+++ b/gpu-operator/getting-started.rst
@@ -406,6 +406,23 @@ options are used with the container-toolkit deployed with GPU Operator:
       - name: CONTAINERD_SET_AS_DEFAULT
         value: true
 
+
+If you need to specify custom values, refer to the following sample command for the syntax:
+
+
+.. code-block:: console
+
+  helm install gpu-operator -n gpu-operator --create-namespace \
+    nvidia/gpu-operator $HELM_OPTIONS \
+      --set toolkit.env[0].name=CONTAINERD_CONFIG \
+      --set toolkit.env[0].value=/etc/containerd/config.toml \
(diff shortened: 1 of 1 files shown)


mikemckiernan:
@geekloper, thank you for the PR.  If I read correctly, this PR adds a sample command that uses the default values.  The Rancher and MicroK8s sections that are after your update are similar, but have the advantage of showing deployment-specific values.  I feel like there's benefit when the example solves a specific challenge.  Please let me know if I'm overlooking some other benefit.

geekloper:
Thanks for your feedback. 
Well, I noticed a documentation gap for Containerd setups and thought it worth mentioning, as setting Helm values can sometimes be tricky. Aimed to mirror clarity provided for MicroK8s and Rancher. If it doesn't align with documentation goals, i’m okay with closing the PR. PR 😉

gpu-operator/getting-started.rst
```diff
@@ -406,6 +406,21 @@ options are used with the container-toolkit deployed with GPU Operator:
       - name: CONTAINERD_SET_AS_DEFAULT
         value: true
 
+These options can be passed to GPU Operator during install time as below.
```
mikemckiernan, line 409: I agree that the syntax for the options can be tricky.  Do you mind revising the intro statement so the purpose is to clarify the syntax?  Something like the foll?

If you need to specify custom values, refer to the following sample command for the syntax:

mikemckiernan:
@geekloper, can you squash your commits and sign?

geekloper:
@mikemckiernan it's done ✅

mikemckiernan:
Thank you again @geekloper!

PR #37, opened 2024-04-10
Bash install for nvaie v5.0

mikemckiernan:


ArangoGutierrez:
Could you provide some more context, copy from where?

mikemckiernan:
> Could you provide some more context, copy from where?

Sorry Eduardo.  My understanding is that the NVAIE doc strategy is in flux and Francis asked me to copy the content from the foll URL into our docs.  It's a curious approach to me, but I yielded to the request to give NVAIE the time and space to work out a more compelling approach across all software components that is uses.

https://docs.nvidia.com/ai-enterprise/latest/user-guide/index.html#installing-nvidia-gpu-operator-with-bash-shell-script

PR #58, opened 2024-05-13
Trim Container Toolkit menu

mikemckiernan:
@elezar, rather than backport the GDPR footer updates (and deal with builder container issues), I'd like to trim the menu.


a-mccarthy:
close this as we've updated the theme and have moved this information to a different versions file and have started trimming it to a smaller version set already.

PR #59, opened 2024-05-13
Add canonical URL to head

mikemckiernan:


a-mccarthy:
hey @mikemckiernan looks like some of this is outdated, but overall still a valuable thing to add as baseurl. With the recent template updates, do we just need to add the baseurl option into the `repo.toml` file?

mikemckiernan:
> hey @mikemckiernan looks like some of this is outdated, but overall still a valuable thing to add as baseurl. With the recent template updates, do we just need to add the baseurl option into the `repo.toml` file?

IIRC, this PR was when I learned that repo_docs > 0.39 is no longer publicly installable.  To mimic it, we'd need to put switch-like code in the `sphinx_conf_py_extra` section to set `html_base_url` according to which docset is building.  Right after I puzzled through that, I realized that Google has no problem finding the documentation and abandoned further plans--but failed to tidy up by closing this PR.  If you want to pursue it, I can assist--if you didn't follow what I described.

a-mccarthy:
@mikemckiernan I'm good with closing this if its no longer needed. thanks!

PR #62, opened 2024-05-17
Unlisted distributions

elezar:


container-toolkit/supported-platforms.md
```diff
@@ -16,7 +16,12 @@ backlinks: none
 
 ## Linux Distributions
 
-Supported Linux distributions are listed below:
+The NVIDIA Container Toolkit has been qualified against the following platforms.
```
mikemckiernan, line 19: I suspect your choice of "qualified" rather than supported is to communicate some subtlety.  If feels dicey to me, but it's you who answers questions and not me.

nit: s/has been/is/, unless there is a consideration that I'm overlooking.

container-toolkit/supported-platforms.md
```diff
@@ -16,7 +16,12 @@ backlinks: none
...
+
+Note that the installation packages are generally forward-compatible and the
+omission of newer distributions from this list below does not indicate that the
+NVIDIA Container Toolkit cannot be used there. Specific issues with newer
+distributions should be reported.
```
mikemckiernan, line 24: I think this can be rephrased more concisely.

```
NVIDIA Container Toolkit is generally forward compatible and works successfully on newer versions of the operating systems in the following table.
```

I've puzzled through the thought of directing readers to report issues from this section of the text, and I can't think why it's more compelling to urge that action here than in any other section.   If you prefer to keep it, consider something like "Open an [issue](https://github.com/NVIDIA/nvidia-container-toolkit/issues) in the toolkit repository to report problems."

mikemckiernan (approved):
The updates appear to increase the scope of support and that's counter to my instincts--but you must have your reasons.  PLMK what I can clarify and thank you.

container-toolkit/supported-platforms.md
elezar: The issue is that people reach out to us all the time pointing out that we need to add "Debian 12" support or "Ubuntu 24.04" support, where we are reasonably confident that the packages should work as is. The caveat is that we have not officiallty tested against these configurations. If you have a suggestion on how to better call this out, that would be appreciated.

elezar: For example: https://github.com/NVIDIA/nvidia-container-toolkit/issues/482#issuecomment-2122204585

jgehrcke: This is a great and important discussion. I wonder: what do we really do?

I see a contradiction between

> we have not officially tested against these configurations

and

> has been qualified

I see "qualification testing" as one of the later stages of testing in a release process, in particular meaning: I take a release candidate and make sure it works against all supported platforms.

I can see that we do _not_ do that yet, and that would be perfectly fine.

In absence of massive release qualification / testing resources it is understandable that _guarantees_ should not be expressed with strong wording.

Towards the community, however, it's typically important and sufficient to express the _intent_ in a friendly tone. That removes much attack surface / source of frustration.

I will work on a proposal!

jgehrcke:
Great discussion. I thin that after #159 we can close this. But this is certainly an area where we can and should do further optimization in the future.

PR #77, opened 2024-07-27
Support for COS

mikemckiernan:
RN: https://nvidia.github.io/cloud-native-docs/review/pr-77/gpu-operator/latest/release-notes.html#new-features

Review HTML: https://nvidia.github.io/cloud-native-docs/review/pr-77/gpu-operator/latest/google-gke.html

### DIFF
--- a/gpu-operator/google-gke.rst
+++ b/gpu-operator/google-gke.rst
@@ -30,21 +30,21 @@ NVIDIA GPU Operator with Google GKE
 About Using the Operator with Google GKE
 ****************************************
 
-You can use the NVIDIA GPU Operator with Google Kubernetes Engine (GKE),
-but you must use an operating system that is supported by the Operator.
+There are two ways to use NVIDIA GPU Operator with Google Kubernetes Engine (GKE):
 
-By default, Google GKE configures nodes with the Container-Optimized OS with Containerd from Google.
-This operating system is not supported by the Operator.
+- Use the default operating system, Container-Optimized OS.
 
-To use a supported operating system, such as Ubuntu 22.04 or 20.04, configure your
-GKE cluster entirely with Ubuntu containerd nodes images or with a node pool
-that uses Ubuntu containerd node images.
+  GKE configures the nodes with containerd from Google and the Google driver installer manages the NVIDIA GPU Driver.
+  The GPU Operator manages the lifecycle and upgrade of the remaining software components such as the device plugin, container toolkit, DCGM, and so on.
 
-By selecting a supported operating system rather than Container-Optimized OS with Containerd,
-you can customize which NVIDIA software components are installed by the GPU Operator at deployment time.
-For example, the Operator can deploy GPU driver containers and use the Operator
-to manage the lifecycle of the NVIDIA software components.
+- Use Ubuntu operating system with containerd.
 
+  You can choose to use the Google driver installer to manage the NVIDIA GPU Driver.
+  Alternatively, you can use the Operator and NVIDIA Driver Manager to manage the driver lifecycle and upgrades.
+
+The preceding approaches relate to using GKE Standard node pools.
+For Autopilot Pods, using the GPU Operator is not supported, and you can refer to
+`Deploy GPU workloads in Autopilot <https://cloud.google.com/kubernetes-engine/docs/how-to/autopilot-gpus>`__.
 
 *************
 Prerequisites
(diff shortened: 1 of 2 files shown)


gpu-operator/google-gke.rst
```diff
@@ -67,11 +67,115 @@ Prerequisites
...
+Perform the following steps to create a GKE cluster with the ``gcloud`` CLI and use Google driver installer to manage the GPU driver.
+You can create a node pool that uses a Container-Optimized OS node image or a Ubuntu node image.
+
+#. Create the cluster.
+   Refer to `Running GPUs in GKE Standard clusters <https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#create>`__
+   in the GKE documentation.
+
+   When you create the cluster, specify the following additional ``gcloud`` command-line options:
+
+   - ``--node-labels="gke-no-default-nvidia-gpu-device-plugin=true"``
```
cdesiniotis, line 83: Question -- is this required for Ubuntu node pools? I only ask since our existing documentation, which only covers Ubuntu, does not mention this label.

gpu-operator/google-gke.rst
```diff
@@ -67,11 +67,115 @@ Prerequisites
...
+
+   Using NVIDIA MIG Manager to configure devices is an alternative to the default GKE approach and is incompatible with the default GKE approach.
+   Refer to `Running Multi-Instance GPUs <https://cloud.google.com/kubernetes-engine/docs/how-to/gpus-multi>`__
+   in the GKE documentation for information about the default approach.
+
+***************************
+Using NVIDIA Driver Manager
+***************************
+
+Perform the following steps to create a GKE cluster with the ``gcloud`` CLI and use the Operator and NVIDIA Driver Manager to manage the GPU driver.
```
cdesiniotis, line 178: As a reader, it is not entirely clear to me that this section, `Using NVIDIA Driver Manager`, is only applicable for Ubuntu node pools. It may be worth adding a sentence here stating that this approach is only supported on Ubuntu.

gpu-operator/google-gke.rst
```diff
@@ -30,21 +30,21 @@ NVIDIA GPU Operator with Google GKE
...
+  The GPU Operator manages the lifecycle and upgrade of the remaining software components such as the device plugin, container toolkit, DCGM, and so on.
 
-By selecting a supported operating system rather than Container-Optimized OS with Containerd,
-you can customize which NVIDIA software components are installed by the GPU Operator at deployment time.
-For example, the Operator can deploy GPU driver containers and use the Operator
-to manage the lifecycle of the NVIDIA software components.
+- Use Ubuntu operating system with containerd.
 
+  You can choose to use the Google driver installer to manage the NVIDIA GPU Driver.
+  Alternatively, you can use the Operator and NVIDIA Driver Manager to manage the driver lifecycle and upgrades.
```
cdesiniotis, line 43: I am questioning whether `NVIDIA Driver Manager` is the term we should use throughout this page. I do not have a better idea at the moment.

cdesiniotis, line 43: From the NVIDIA perspective, we recommend the latter approach on Ubuntu (have GPU Operator manage every software component) -- I am wondering if we should make that more evident here and in the procedure for Ubuntu.

gpu-operator/google-gke.rst
```diff
@@ -67,11 +67,115 @@ Prerequisites
   in the Google Cloud documentation.
 
 
-*********
-Procedure
-*********
+*********************************
+Using the Google Driver Installer
```
cdesiniotis, line 71: In terms of structure, I am wondering if this page would be easier to follow if the section headers aligned with the bullets from the introduction:
```
Using Container-Optimized OS
...

Using Ubuntu
...
```
For each OS, we provide the recommended procedure. For COS, that would be Google's driver installer + GPU Operator for the rest. For Ubuntu, that would be GPU Operator for everything. 

One downside -- if we wanted to document multiple options for each OS, i.e. documenting two procedure for Ubuntu (Google vs NVIDIA managed drivers), that would lead to some duplication across the various procedures.

cdesiniotis (commented):
Thanks @mikemckiernan, this is a great start. I left a few comments / questions. My main question concerns the overall structure of the document and whether using the operating systems (COS and Ubuntu) as section headers would improve readability. Happy to discuss this.

Dragoncell:
I think the second section `Using NVIDIA Driver Manager` most of part is duplicated to the part `Using Google Driver Installer` 

Maybe combine the command steps, and separate the different steps towards the two approach 


Up to step 7 in `Using Google Driver Installer` are same for `Using NVIDIA Driver Manager` except the Ubuntu node image. 

then for 
1.  `Using Google Driver Installer` : 1. install Google Driver installer  2. Install Operator 
2.  `Using NVIDIA Driver Manager` : 1. Install Operator with driver enabled

gpu-operator/google-gke.rst
cdesiniotis: We discussed this offline and decided the current structure will suffice.

mikemckiernan:
> I think the second section `Using NVIDIA Driver Manager` most of part is duplicated to the part `Using Google Driver Installer`
> 
> Maybe combine the command steps, and separate the different steps towards the two approach
> 
> Up to step 7 in `Using Google Driver Installer` are same for `Using NVIDIA Driver Manager` except the Ubuntu node image.
> 
> then for
> 
> 1. `Using Google Driver Installer` : 1. install Google Driver installer  2. Install Operator
> 2. `Using NVIDIA Driver Manager` : 1. Install Operator with driver enabled

The redundancy is not ideal, but I didn't think they were identical.  The first step in the first procedure is to create a node pool. The first step in the second procedure is to create a cluster.  I also wasn't under the impression that using a ubuntu_containerd image would require applying the node labels or disabling the automatic driver installation.  (Has that changed and the procedure is stale? We can address offline.)

So, based on my understanding that customers either start by creating a node pool with specific node labels and command-line args that aren't common, a middle-of-procedure step to install the driver installer daemon set, and differing arguments to Helm at the end to install the Operator...I felt that was more cognitive load than necessary and could be reduced by separate procedures.

mikemckiernan:
> Thanks @mikemckiernan, this is a great start. I left a few comments / questions. My main question concerns the overall structure of the document and whether using the operating systems (COS and Ubuntu) as section headers would improve readability. Happy to discuss this.

 I started off by using the OS to organize the info, but because COS and Ubuntu are both supported by the Google driver installer, I felt like it got too messy too quickly.  I revised the intro section and I hope it helps orient readers for the decision to make--driver installer or Operator, and more prominently shows the supported OSes.

gpu-operator/google-gke.rst
```diff
@@ -67,11 +86,112 @@ Prerequisites
...
+*********************************
+
+Perform the following steps to create a GKE cluster with the ``gcloud`` CLI and use Google driver installer to manage the GPU driver.
+You can create a node pool that uses a Container-Optimized OS node image or a Ubuntu node image.
+
+#. Create the cluster.
+   Refer to `Running GPUs in GKE Standard clusters <https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#create>`__
+   in the GKE documentation.
+
+   When you create the cluster, specify the following additional ``gcloud`` command-line options:
```
cdesiniotis, line 100: Shouldn't this be `When you create the GPU node pool, ... `?

bobbypage: +1, it should be about node pool

gpu-operator/google-gke.rst
bobbypage: cc @Dragoncell 

I would expect we need it for Ubuntu as well (otherwise we will run the GKE GPU device plugin on those nodes).

gpu-operator/google-gke.rst
```diff
@@ -67,11 +86,112 @@ Prerequisites
...
+
+   - ``--node-labels="gke-no-default-nvidia-gpu-device-plugin=true"``
+
+     The node label disables the GKE GPU device plugin daemon set on GPU nodes.
+
+   - ``--accelerator type=...,gpu-driver-version=disabled``
+
+     This argument prevents automatically installing the GPU driver on GPU nodes.
```
bobbypage, line 108: How about

```
This argument skips automatic GPU driver installation since it's not supported when using the NVIDIA GPU operator.
```

bobbypage, line 108: How about?

```suggestion
     This argument skips automatic GPU driver installation since it's not supported when using the NVIDIA GPU operator.
```

gpu-operator/google-gke.rst
```diff
@@ -67,11 +86,112 @@ Prerequisites
...
+
+#. Get the authentication credentials for the cluster:
+
+   .. code-block:: console
+
+      $ USE_GKE_GCLOUD_AUTH_PLUGIN=True \
```
bobbypage, line 114: I think we can drop the `USE_GKE_GCLOUD_AUTH_PLUGIN=True` and rely on gcloud default, it shouldn't be needed 

```suggestion
     
```

cc @Dragoncell do you have some context on why we suggested this originally?

gpu-operator/google-gke.rst
```diff
@@ -67,11 +86,112 @@ Prerequisites
...
+          gcloud container clusters get-credentials demo-cluster --zone us-west1-a
```
bobbypage, line 115: small nit, we recommend users to avoid zonal clusters, and instead use regional clusters in general.

So I would perhaps put `--location=us-west1` instead.

gpu-operator/google-gke.rst
```diff
@@ -67,11 +86,112 @@ Prerequisites
...
+          -n gpu-operator \
+          nvidia/gpu-operator \
+          --set hostPaths.driverInstallDir=/home/kubernetes/bin/nvidia \
+          --set toolkit.installDir=/home/kubernetes/bin/nvidia \
+          --set cdi.enabled=true \
+          --set cdi.default=true \
+          --set driver.enabled=false
+
+   Set the NVIDIA Container Toolkit and driver installation path to ``/home/kubernetes/bin/nvidia``.
+   On GKE node images, the ``/home`` directory is writable and is a stateful location for storing the NVIDIA runtime binaries.
```
bobbypage, line 180:
```
On GKE node images, the ``/home`` directory is writable and is a stateful location for storing the NVIDIA runtime binaries.
```

This isn't true for all of `/home` in COS. I would remove this paragraph, as it's an implementation detail.

gpu-operator/google-gke.rst
```diff
@@ -160,4 +275,6 @@ Related Information
 
 * If you have an existing GKE cluster, refer to
   `Add and manage node pools <https://cloud.google.com/kubernetes-engine/docs/how-to/node-pools>`_
-  in the Google Kubernetes Engine documentation.
\ No newline at end of file
+  in the GKE documentation.
+* When you create new node pools, specify the ``--node-labels="gke-no-default-nvidia-gpu-device-plugin=true"`` CLI argument
```
bobbypage, line 279: ... and let's also mention to disable the automatic GPU driver installation, i.e.
```
--accelerator type=...,gpu-driver-version=disabled``
```

bobbypage:
Thanks for updating the docs, left a few suggestions!

PR #83, opened 2024-07-31
Clarify when to generate CDI

mikemckiernan:
Review HTML: https://nvidia.github.io/cloud-native-docs/review/pr-83/container-toolkit/latest/cdi-support.html#procedure

### DIFF
--- a/container-toolkit/cdi-support.md
+++ b/container-toolkit/cdi-support.md
@@ -39,8 +39,12 @@ CDI also improves the compatibility of the NVIDIA container stack with certain f
 
 ### Procedure
 
-Two typical locations for CDI specifications are `/etc/cdi/` and `/var/run/cdi`.
-However, the path to create and use can depend on the container engine that you use.
+Two common locations for CDI specifications are `/etc/cdi/` and `/var/run/cdi/`.
+NVIDIA recommends using the `/etc/cdi/` directory for user-applied changes, such as performing the following steps.
+The `/var/run/cdi/` directory is commonly used for dynamic changes made by software.
+The contents of the `/var/run/cdi/` directory can change at runtime and can be deleted after a reboot.
+
+However, the `/etc/cdi/` path to create and use can depend on the container engine that you use.
 
 1. Generate the CDI specification file:
 
--- a/container-toolkit/cdi-support.md
+++ b/container-toolkit/cdi-support.md
@@ -77,7 +77,11 @@ However, the path to create and use can depend on the container engine that you 
    ```
 
 ```{important}
-If you change the device or CUDA driver configuration, you must generate a new CDI specification.
+You must generate a new CDI specification after any of the following changes:
+
+- You change the device or CUDA driver configuration.
+- You upgrade the Linux kernel.
+
 A configuration change can occur when MIG devices are created or removed, or when the driver is upgraded.
 ```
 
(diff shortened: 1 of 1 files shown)


container-toolkit/cdi-support.md
```diff
@@ -77,7 +77,11 @@ However, the path to create and use can depend on the container engine that you
    ```
 
 ```{important}
-If you change the device or CUDA driver configuration, you must generate a new CDI specification.
+You must generate a new CDI specification after any of the following changes:
+
+- You change the device or CUDA driver configuration.
+- You upgrade the Linux kernel.
```
elezar, line 83: There isn't anything kernel-specific in the CDI specification. It would be good to understand why this would be required?

mikemckiernan: Glad you looked and clarified.  I'm OK with leaving this open for a few days and revising if we learn what the issue was.  If we never learn, then you or I can close it.  LMK if you prefer otherwise.

mikemckiernan: @elezar, I replaced the advice based on a sentence you shared in the nvbug. PTAL [again](https://nvidia.github.io/cloud-native-docs/review/pr-83/container-toolkit/latest/cdi-support.html#procedure).

Suggs and corrections welcome, as always.

container-toolkit/cdi-support.md
```diff
@@ -39,8 +39,12 @@ CDI also improves the compatibility of the NVIDIA container stack with certain f
...
 ### Procedure
 
-Two typical locations for CDI specifications are `/etc/cdi/` and `/var/run/cdi`.
-However, the path to create and use can depend on the container engine that you use.
+Two common locations for CDI specifications are `/etc/cdi/` and `/var/run/cdi/`.
+NVIDIA recommends using the `/etc/cdi/` directory for user-applied changes, such as performing the following steps.
+The `/var/run/cdi/` directory is commonly used for dynamic changes made by software.
+The contents of the `/var/run/cdi/` directory can change at runtime and can be deleted after a reboot.
+
+However, the `/etc/cdi/` path to create and use can depend on the container engine that you use.
```
elezar, line 47: It's not just `/etc/cdi` that is configuration dependent, so I prefer something a little closer to the original text.

container-toolkit/cdi-support.md
elezar: Let's remove the kernel upgrade text. What about something along the lines of:

> You must generate a new CDI specification if the device or CUDA driver configuration changes. In addition, if a location such as `/var/run/cdi` -- which is cleared on boot -- is used, the CDI specifications must be regenerated at boot.

container-toolkit/cdi-support.md
```diff
@@ -39,8 +39,12 @@ CDI also improves the compatibility of the NVIDIA container stack with certain f
 
 ### Procedure
 
-Two typical locations for CDI specifications are `/etc/cdi/` and `/var/run/cdi`.
-However, the path to create and use can depend on the container engine that you use.
+Two common locations for CDI specifications are `/etc/cdi/` and `/var/run/cdi/`.
+NVIDIA recommends using the `/etc/cdi/` directory for user-applied changes, such as performing the following steps.
```
elezar, line 43: I don't know whether I want to make this recommendation at this stage. I can see us moving to using `/var/run/cdi` by default as soon as we have tooling to generate these specs at boot.

container-toolkit/cdi-support.md
```diff
@@ -39,8 +39,12 @@ CDI also improves the compatibility of the NVIDIA container stack with certain f
...
+The `/var/run/cdi/` directory is commonly used for dynamic changes made by software.
+The contents of the `/var/run/cdi/` directory can change at runtime and can be deleted after a reboot.
```
elezar, line 45: Using "can be deleted" here seems to imply an action by the user. It is also not clear what "can change at runtime" implies.

elezar (changes requested):
Thanks for taking another pass at this @mikemckiernan. I need to think on better wording here. Let me get back to you.

container-toolkit/cdi-support.md
mikemckiernan: Very sorry to have accidentally left the kernel upgrade mention.

PR #86, opened 2024-08-08
update nvidia-ctk version in the rhel install section

tariq1890:


gpu-operator/getting-started.rst
```diff
@@ -325,9 +325,9 @@ In this scenario, use the NVIDIA Container Toolkit image that is built on UBI 8:
    $ helm install --wait --generate-name \
         -n gpu-operator --create-namespace \
         nvidia/gpu-operator \
-        --set toolkit.version=v1.15.0-ubi8
+        --set toolkit.version=v1.16.1-ubi8
 
-Replace the ``v1.15.0`` value in the preceding command with the version that is supported
+Replace the ``v1.16.1`` value in the preceding command with the version that is supported
 with the NVIDIA GPU Operator.
 Refer to the :ref:`GPU Operator Component Matrix` on the platform support page.
```
mikemckiernan, line 332: I suspect we don't need to update these values with each release and can rely on customers to review the support matrix.  But I also have no reason to object to updating it. TY!

mikemckiernan (approved):

PR #94, opened 2024-08-30
Info from Francis

mikemckiernan:
Review HTML: https://nvidia.github.io/cloud-native-docs/review/pr-94/gpu-operator/latest/platform-support.html#supported-operating-systems-and-kubernetes-platforms

### DIFF
--- a/gpu-operator/platform-support.rst
+++ b/gpu-operator/platform-support.rst
@@ -311,8 +311,8 @@ The GPU Operator has been validated in the following scenarios:
 
        * - | Red Hat
            | Enterprise
-           | Linux 8.4,
-           | 8.6---8.9
+           | Linux 8.6,
+           | 8.8, 8.10
          - 1.22---1.30
          -
          -


mikemckiernan:
@francisguillier, PTAL about the RHEL 8.10 update when you get a chance.

francisguillier:
yes, looks very good.

annex: I need to figure out the situation for HPE Ezmeral Runtime Enterprise - but we can publish this PR.

thanks @mikemckiernan !

francisguillier:
@mikemckiernan 
I just realize we need to make the change here as well:
https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/platform-support.html#supported-operating-systems-and-kubernetes-platforms

"Red Hat
Enterprise
Linux 8.4,
8.6—8.9"

should be replaced with

"Red Hat
Enterprise
Linux 8.6,
8.8, 8.10"

PR #98, opened 2024-09-10
Troubleshoot nouveau

mikemckiernan:
Review HTML: https://nvidia.github.io/cloud-native-docs/review/pr-98/gpu-operator/latest/troubleshooting.html#gpu-operator-validator-failed-to-create-pod-sandbox

@francisguillier, PTAL.


francisguillier:
@mikemckiernan 
the content is good but I think the NVEX team wanted a note in the install guide to mention that NOUVEAU must be disabled on the nodes before installing GPU operator. Otherwise the driver container deployment will fail.

PR #114, opened 2024-10-18
Update BM and VMs to OCP 4.17

mikemckiernan:
@francisguillier, PTAL.  Is this table the only update?  

https://nvidia.github.io/cloud-native-docs/review/pr-114/gpu-operator/latest/platform-support.html#supported-operating-systems-and-kubernetes-platforms

We don't update the table a little further down for virtualization?

https://nvidia.github.io/cloud-native-docs/review/pr-114/gpu-operator/latest/platform-support.html#support-for-kubevirt-and-openshift-virtualization

TY!


francisguillier:
yes you are right, we should update it too in the virtualization table.
the change you did on the other tables look good to me.

PR #126, opened 2024-11-08
Add ubuntu24.04 to list of supported platforms

elezar:


mikemckiernan (approved):
LGTM, TY as always!

mikemckiernan:
I tagged and pushed, btw.

PR #133, opened 2024-12-03
Specify namespace when getting GPU driver upgrade events

cdesiniotis:
Events are namespace-scoped. The GPUDriverUpgrade events are associated with Node objects and are emitted in the default namespace. This aligns with the Node events generated by the kubelet, which also show up in the default namespace.


mikemckiernan (approved):
That's valuable information, TY!

PR #137, opened 2024-12-16
Update platform-support.rst

a-mccarthy:


mikemckiernan:
Thanks for the fixups on the VMware typos!  LGTM.

All the cloud-native projects require submitters to sign their commits--that's the DCO check.  My recollection is that I used [GPG to create a key](https://docs.github.com/en/authentication/managing-commit-signature-verification/generating-a-new-gpg-key), but it has been a while.

a-mccarthy:
thanks @mikemckiernan! I'm working on getting my key set up for signing so i can fix the commit here

mikemckiernan (approved):
LGTM, thanks again!

PR #139, opened 2024-12-22
Fix the option path for disabling OpenShift Driver Toolkit

empovit:


a-mccarthy (approved):
looks good to me, thanks @empovit

PR #142, opened 2025-01-17
Update theme

mikemckiernan:
This is a doozie to review.  I suppose I can break it up into smaller parts if you find it a hassle to review.


a-mccarthy:
Talked with @mikemckiernan about this PR. I think the changes look good. 

Some notes on the changes:
* The version picker is not working in the preview b/c it relies on a change to an internal repo that helps to build/publish the docs. This _should_ not be an issue on the live site once everything is merged.
* Updates to using a version1.json file instead of the current version.json file to align with the theme requirements. When we release a new version we only need to update the version1.json file for the new component.
* The code block to create a table of contents is no longer needed for pages. This is handled automatically by the theme.

PR #151, opened 2025-01-29
Align container tags

mikemckiernan:


.github/workflows/docs-build.yaml
```diff
@@ -10,7 +10,7 @@ on:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
-  TAG: 0.3.0
+  TAG: 0.4.0
```
mikemckiernan, line 13: I need to update this tag so that when it's merged the build noticed that `0.4.0` does not exist and builds the `0.4.0` image that has the updated theme and so on.

a-mccarthy (approved):

.gitlab-ci.yml
```diff
@@ -1,7 +1,7 @@
 variables:
   CONTAINER_TEST_IMAGE: "${CI_REGISTRY_IMAGE}:${CI_COMMIT_REF_SLUG}"
   CONTAINER_RELEASE_IMAGE: "${CI_REGISTRY_IMAGE}:0.5.0"
-  BUILDER_IMAGE: ghcr.io/nvidia/cloud-native-docs:0.2.0
+  BUILDER_IMAGE: ghcr.io/nvidia/cloud-native-docs:0.4.0
```
mikemckiernan, line 4: This needs to match the ghcr.io image so that the HTML is identical.

mikemckiernan (commented):
PLMK if I can clarify anything or if anything is concerning.

PR #152, opened 2025-01-30
remove Debian 10/CentOS7/RHEL7/Ubuntu18 from container toolkit supported platforms

a-mccarthy:


elezar (approved):

jgehrcke:
Let's always take a few seconds and summarize the motivation/reason for a change! :) Thanks!

PR #158, opened 2025-02-13
toolkit: install guide: refer to section about supported platforms, tweak headings

jgehrcke:
I find it important that the installation section informs about supported platforms. We have a section for that (in a different doc) so all we are missing is a cross-reference. 

Is Sphinx our documentation generator? I think / hope that cross-references in md-sphinx work as I am trying here, but I did not test this locally.

Also tweaking the headings to shift attention to the distribution names more than the tooling.

### DIFF
--- a/container-toolkit/install-guide.md
+++ b/container-toolkit/install-guide.md
@@ -10,14 +10,14 @@
 
 ### Prerequisites
 
-Install the NVIDIA GPU driver for your Linux distribution.
+1. Install the NVIDIA GPU driver for your Linux distribution.
 NVIDIA recommends installing the driver by using the package manager for your distribution.
-
 For information about installing the driver with a package manager, refer to
 the [_NVIDIA Driver Installation Quickstart Guide_](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html).
+Alternatively, you can install the driver by [downloading](https://www.nvidia.com/en-us/drivers/) a `.run` installer.
 
-Alternatively, you can install the driver by downloading a `.run` installer.
-Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.aspx?lang=en-us) page.
+
+2. Read [this section](./supported-platforms.md) about platform support.
 
 ### Debian-based distributions (Ubuntu):
 
--- a/container-toolkit/install-guide.md
+++ b/container-toolkit/install-guide.md
@@ -103,11 +103,6 @@ Alternatively, you can install the driver by [downloading](https://www.nvidia.co
    $ sudo zypper modifyrepo --enable nvidia-container-toolkit-experimental
    ```
 
-   <!--
-   TODO:
-   - [ ] Experimental repos: zypper modifyrepo --enable nvidia-container-toolkit-experimental
-   -->
-
 1. Install the NVIDIA Container Toolkit packages:
 
    ```console
--- a/container-toolkit/install-guide.md
+++ b/container-toolkit/install-guide.md
@@ -19,7 +19,7 @@ the [_NVIDIA Driver Installation Quickstart Guide_](https://docs.nvidia.com/data
 Alternatively, you can install the driver by downloading a `.run` installer.
 Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.aspx?lang=en-us) page.
 
-### Installing with Apt
+### Debian-based distributions (Ubuntu):
 
 1. Configure the production repository:
 
@@ -53,7 +53,7 @@ Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.asp
    $ sudo apt-get install -y nvidia-container-toolkit
    ```
 
-### Installing with Yum or Dnf
+### RPM-based distributions (RHEL/CentOS, Fedora, Amazon Linux, ...):
 
 1. Configure the production repository:
 
@@ -74,7 +74,7 @@ Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.asp
    $ sudo yum install -y nvidia-container-toolkit
    ```
 
-### Installing with Zypper
+### OpenSUSE/SLE:
 
 1. Configure the production repository:
 
--- a/container-toolkit/install-guide.md
+++ b/container-toolkit/install-guide.md
@@ -4,7 +4,6 @@
 
 # Installing the NVIDIA Container Toolkit
 
-
 ## Installation
 
 (pre-requisites)=
@@ -19,6 +18,8 @@ For information about installing the driver with a package manager, refer to
 the [_NVIDIA Driver Installation Quickstart Guide_](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html).
 Alternatively, you can install the driver by [downloading](https://www.nvidia.com/en-us/drivers/) a `.run` installer.
 
+
+(installing-with-apt)=
 
 ### With `apt`: Ubuntu, Debian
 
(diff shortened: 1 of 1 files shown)


jgehrcke:
Sweeet! :balloon: 
```
Artifact html-build-artifact.zip successfully finalized. Artifact ID 2585643272
Artifact html-build-artifact has been successfully uploaded! Final size is 34511668 bytes. Artifact ID is 2585643272
Artifact download URL: https://github.com/NVIDIA/cloud-native-docs/actions/runs/13308323093/artifacts/2585643272
```
Looking at that.

container-toolkit/install-guide.md
```diff
@@ -10,16 +10,16 @@
...
 NVIDIA recommends installing the driver by using the package manager for your distribution.
-
 For information about installing the driver with a package manager, refer to
 the [_NVIDIA Driver Installation Quickstart Guide_](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html).
+Alternatively, you can install the driver by [downloading](https://www.nvidia.com/en-us/drivers/) a `.run` installer.
+
 
-Alternatively, you can install the driver by downloading a `.run` installer.
-Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.aspx?lang=en-us) page.
+2. Read [this section](./supported-platforms.md) about platform support.
```
elezar, line 20: Nit: Should this be first?

container-toolkit/install-guide.md
```diff
@@ -10,16 +10,16 @@
...
 
-### Installing with Apt
+### Debian-based distributions (Ubuntu):
```
elezar, line 22: Does this read as if we're only targetting Ubuntu?
```suggestion
### Debian-based distributions (e.g. Ubuntu):
```

container-toolkit/install-guide.md
```diff
@@ -53,7 +53,7 @@ Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.asp
    $ sudo apt-get install -y nvidia-container-toolkit
    ```
 
-### Installing with Yum or Dnf
+### RPM-based distributions (RHEL/CentOS, Fedora, Amazon Linux, ...):
```
elezar, line 56:
```suggestion
### RPM-based distributions (e.g. RHEL/CentOS, Fedora, Amazon Linux, ...):
```

container-toolkit/install-guide.md
```diff
@@ -74,7 +74,7 @@ Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.asp
    $ sudo yum install -y nvidia-container-toolkit
    ```
 
-### Installing with Zypper
+### OpenSUSE/SLE:
```
elezar, line 77: It's a little strange that this is also RPM-based but has its own section. Should we call that out above or have subsections for the two package manager types?

container-toolkit/install-guide.md
a-mccarthy: +1 i think it should be first as well

and an additional nit to embed the text into the link, so something like 
``` Make sure you are using a [supported platform](./supported-platforms.md).```

Comment (jgehrcke) [reply container-toolkit/install-guide.md:20]: > Nit: Should this be first?

Yes, I was hoping for that feedback!

Comment (jgehrcke) [reply container-toolkit/install-guide.md:20]: >  something like Make sure you are using a [supported platform](./supported-platforms.md).

In general I like this approach. Here I opted for a subtly different approach because:

It's really a section about platform support in general, and we precisely now want to be mindful to _not_ communicate something in the lines of "do not continue if you do not use a strictly supported platform" :).

Related:   https://github.com/NVIDIA/cloud-native-docs/pull/62 and e.g. https://github.com/NVIDIA/nvidia-container-toolkit/issues/482#issuecomment-2122204585.

File: container-toolkit/install-guide.md (hunk 2 above)
Comment (jgehrcke) [reply container-toolkit/install-guide.md:22]: Right, I forgot to add ", ..." like I did below. In this case I think I like this a little better than "e.g."

File: container-toolkit/install-guide.md (hunk 4 above)
Comment (jgehrcke) [reply container-toolkit/install-guide.md:77]: Yes you have a strong point here. Will think about that.

File: container-toolkit/install-guide.md (hunk 3 above)
Comment (jgehrcke) [reply container-toolkit/install-guide.md]: I started to overthink this. Removed the "RPM-based distributions" again.

Pushed a new commit, let's see how this looks when its rendered! :) I like to first-class the distribution names instead of the tooling names (that was my main angle here). There's no perfect solution though.

Some inspiration from..


### chrome

image

### vscode

image


### gcloud cli

image

File: container-toolkit/install-guide.md (hunk 2 above)
Comment (jgehrcke) [reply container-toolkit/install-guide.md]: Now I have clarity: we don't need generic enumerations. In these headings, we want to make sure we list all those distros that are in the table of "supported platforms". That's all we need. Pushed a commit.

File: container-toolkit/install-guide.md (hunk 4 above)
Comment (jgehrcke) [reply container-toolkit/install-guide.md]: Happy with the current state: https://nvidia.github.io/cloud-native-docs/review/pr-158/container-toolkit/latest/install-guide.html.

File: container-toolkit/install-guide.md [hunk 5]
```diff
@@ -4,22 +4,23 @@
...
+NVIDIA recommends installing the driver by using the package manager for your distribution.
 For information about installing the driver with a package manager, refer to
 the [_NVIDIA Driver Installation Quickstart Guide_](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html).
+Alternatively, you can install the driver by [downloading](https://www.nvidia.com/en-us/drivers/) a `.run` installer.
 
-Alternatively, you can install the driver by downloading a `.run` installer.
-Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.aspx?lang=en-us) page.
 
-### Installing with Apt
+### Ubuntu and Debian
```
elezar, line 23: I'm happy with this change for ubuntu and debian since I think this covers most users of `apt`. My concern is that using the same construct for RPM-based distributions where there are other prominent ones that are not listed. (*Rocky* comes to mind). This will definitely lead to GitHub issues along the lines of "RockyLinux support when?". At present my response is to link the "Installing with Yum or Dnf" section, but I can see more questions being raised if I link a section that's titled "RHEL/CentOS, Fedora, Amazon Linux".

jgehrcke: yes. note: added a commit after working on this with Evan

jgehrcke:
Pushed more commits after iterating on this with Evan. Will want to double-check the next rendered version before merge. Will report back.

jgehrcke:
Good: the legacy anchors now work. Example: https://nvidia.github.io/cloud-native-docs/review/pr-158/container-toolkit/latest/install-guide.html#installing-with-apt

elezar (approved):
I'm happy with these. Thanks @jgehrcke

container-toolkit/install-guide.md
```diff
@@ -10,16 +10,22 @@
...
+1. Read [this section](./supported-platforms.md) about platform support.
 
+2. Install the NVIDIA GPU driver for your Linux distribution.
+NVIDIA recommends installing the driver by using the package manager for your distribution.
 For information about installing the driver with a package manager, refer to
 the [_NVIDIA Driver Installation Quickstart Guide_](https://docs.nvidia.com/datacenter/tesla/tesla-installation-notes/index.html).
+Alternatively, you can install the driver by [downloading](https://www.nvidia.com/en-us/drivers/) a `.run` installer.
 
-Alternatively, you can install the driver by downloading a `.run` installer.
-Refer to the NVIDIA [Official Drivers](https://www.nvidia.com/Download/index.aspx?lang=en-us) page.
```
a-mccarthy, line 20: Is there a reason why we got ride of this link? I dont remember if this has come up before in a previous comments. thanks

jgehrcke: No removal. I replaced it with a more modern URL. You will notice in your browser that the old URL redirects to `/en-us/drivers`.

Personally, I like to look at changes like this in the rendered view.

We should also work towards using often called "semantic line breaks" to keep the markdown source diff better reviewable.

jgehrcke:
Thanks for discussion, let's certainly iterate more soon!

PR #159, opened 2025-02-13
toolkit: supported platforms: note that things may work on other/newer distros

jgehrcke:
I was reading through the discussion in https://github.com/NVIDIA/cloud-native-docs/pull/62 and found it very tangible. Instead of iterating on the PR I thought I create a new one with my proposal to address some of the points discussed there.

I very much agree with @elezar that we need a note about those distributions that are _not_ listed in that table, and that we want to invite people to just try things out. This is the main part of this patch.

Additional changes:

- Centos -> CentOS
- Instead of loading lots of meaning into a single word: ["supported" vs "qualified"](https://github.com/NVIDIA/cloud-native-docs/pull/62#discussion_r1606776018) I propose using slightly more exhaustive wording: "qualification-tested and expected to work on these platforms"
- We did use footnotes, and I have changed one note about Tegra into a footnote.

### DIFF
--- a/container-toolkit/supported-platforms.md
+++ b/container-toolkit/supported-platforms.md
@@ -4,29 +4,39 @@
 
 (supported-platforms)=
 
-# Supported Platforms
+# Platform support
+
+Recent NVIDIA Container Toolkit releases are tested and expected to work on these Linux distributions:
+
+| OS Name / Version        | amd64 / x86_64 | ppc64le | arm64 / aarch64 {sup}`1` |
+| ------------------------ | -------------- | ------- | ------------------------ |
+| Amazon Linux 2023        | X              |         | X {sup}`2`               |
+| Amazon Linux 2           | X              |         | X                        |
+| Open Suse/SLES 15.x      | X              |         |                          |
+| Debian Linux 11          | X              |         |                          |
+| CentOS 8                 | X              | X       | X                        |
+| RHEL 8.x                 | X              | X       | X                        |
+| RHEL 9.x                 | X              | X       | X                        |
+| Ubuntu 20.04             | X              | X       | X                        |
+| Ubuntu 22.04             | X              | X       | X                        |
+| Ubuntu 24.04             | X              |         | X                        |
 
 
-## Linux Distributions
+## Please report issues
 
-Supported Linux distributions are listed below:
+Our qualification-testing procedures are constantly evolving and we might miss
+certain problems. Please
+[report](https://github.com/NVIDIA/nvidia-container-toolkit/issues) issues in
+particular as they occur on a platform listed above.
 
-| OS Name / Version        | amd64 / x86_64 | ppc64le | arm64 / aarch64 |
-| ------------------------ | -------------- | ------- | --------------- |
-| Amazon Linux 2023        | X              |         | X {sup}`1`      |
-| Amazon Linux 2           | X              |         | X               |
-| Open Suse/SLES 15.x      | X              |         |                 |
-| Debian Linux 11          | X              |         |                 |
-| Centos 8                 | X              | X       | X               |
-| RHEL 8.x                 | X              | X       | X               |
-| RHEL 9.x                 | X              | X       | X               |
-| Ubuntu 20.04             | X              | X       | X               |
-| Ubuntu 22.04             | X              | X       | X               |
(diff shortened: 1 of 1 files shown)


elezar (approved):
I think I'm happy with the changes.

@a-mccarthy any further comments?

container-toolkit/supported-platforms.md
```diff
@@ -9,28 +9,44 @@
...
+| RHEL 7.x                 | X              | X       |                          |
+| RHEL 8.x                 | X              | X       | X                        |
+| RHEL 9.x                 | X              | X       | X                        |
+| Ubuntu 18.04             | X              | X       | X                        |
+| Ubuntu 20.04             | X              | X       | X                        |
+| Ubuntu 22.04             | X              | X       | X                        |
+| Ubuntu 24.04             | X              |         | X                        |
+
+
+Note:
```
a-mccarthy, line 32: Maybe lets put this in its own section? or as some paragraphs after the table. Otherwise the footnotes get pushed down further and we have a bit of note inception going on, notes within notes.

Maybe something like "Using unsupported distros" or "Distribution testing" or "Additional distros" for a section heading. Then we can go into details on the qualification testing done and your note on using unsupported distros how to give us feedback?

container-toolkit/supported-platforms.md
```diff
@@ -9,28 +9,44 @@
...
-| RHEL 9.x                 | X              | X       | X               |
-| Ubuntu 18.04             | X              | X       | X               |
-| Ubuntu 20.04             | X              | X       | X               |
-| Ubuntu 22.04             | X              | X       | X               |
-| Ubuntu 24.04             | X              |         | X               |
-
-The `arm64` / `aarch64` architecture includes support for Tegra-based systems.
-
-1. For Amazon Linux 2023 on Arm64, a `g5g.2xlarge` Amazon EC2 instance was used for validation.
+Releases are qualification-tested and expected to work on these platforms:
```
a-mccarthy, line 12: I'm not sure what qualification-testing means. Is it generic enough that readers would know what it means? Are there qualifications that we test that are specific to the software that we can link to?

I'd be a little cautious with removing the word "support" entirely here. B/c I don't want this to read like "we support everything, everywhere, all the time". Is there an SLA on support and what the community can reasonably expect in terms of getting questions answered for unsupported use cases?

a-mccarthy:
@jgehrcke thanks for the PR! I think its great to encourage the community to try out stuff and give us feedback. It feels like 2 separate concerns here though. 

1. List what is supported and what users can expect to be supported to get questions answered if they run into issues
2. Letting folks know a bit about the testing and that things should work on other distros if they want to experiment or have a need to use something not on the official list

I added a few comments about this inline as well, but I'm thinking it might be best to have your additions in their own section. Just my 2 cents, I'm not overly familiar with the testing and support matrix in depth though, so i might not be thinking about things the same as you all :)

container-toolkit/supported-platforms.md
jgehrcke: > I'm not sure what qualification-testing means.

Qualification-testing IMO in software typically refers to a quality assurance stage as part of a release process, trying to make (pretty damn) sure that things really work, in addition to, say, CI tests.

Other than that, I allow myself to make a bit of a joke here and refer to NASA:

image

I think you inspire me to simplify. I will remove the "qualification-". Simplicity wins, so I will push a commit changing this to:

> Releases are tested and expected to work on these platforms

container-toolkit/supported-platforms.md
jgehrcke: > Maybe lets put this in its own section? or as some paragraphs after the table.

Yes, I like that you say this. Thought about it too, was looking for feedback about this in particular. 

I will add sub sections, one about reporting issues in general, and one about "Other Linux distributions".

jgehrcke:
I pushed an update. This is how this currently renders in my VSCode:

image

jgehrcke:
conflicting with https://github.com/NVIDIA/cloud-native-docs/pull/152

jgehrcke:
Merge conflict resolved, squashed into single commit, worked on commit msg.

jgehrcke:
Going ahead here, we can always discuss and iterate more post-merge. Thanks for feedback, everyone.

Issue #155, opened 2025-02-10
Improve security docs


a-mccarthy:
We should improve our security documentation to make it easier for users to understand the security requirements for the GPU operator. This should be put on a new Security page and include the permissions and requirements of the using GPU operator, security best practices to follow, and a list of security disclosures. 

To Do:

- [ ] Create a new Security page under the NVIDIA GPU Operator heading. Move existing security content to the new page, https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/index.html#pod-security-context-of-the-operator-and-operands
- [ ] Add a place to track existing CVEs for the Operator and its components, link to CVE release page
- [ ] Add links to NVIDIA security disclosure policy and any relevant information related to our security response process
- [ ] Review and revise content in https://github.com/NVIDIA/cloud-native-docs/pull/136, add it to the new security page.

a-mccarthy:
cc: @cdesiniotis and @tariq1890

PR #161, opened 2025-02-18
Add security page

a-mccarthy:
Part of https://github.com/NVIDIA/cloud-native-docs/issues/155

### DIFF
--- a/gpu-operator/security.rst
+++ b/gpu-operator/security.rst
@@ -1,0 +1,99 @@
+
+*****************************
+Security Considerations
+*****************************
+
+
+Pod Security Context of the Operator and Operands
+=================================================
+
+Several of the NVIDIA GPU Operator operands, such as the driver containers and container toolkit,
+require the following elevated privileges:
+
+- ``privileged: true``
+- ``hostPID: true``
+- ``hostIPC: true``
+
+The elevated privileges are required for the following reasons:
+
+- Access to the host file system and hardware devices, such as NVIDIA GPUs.
+- Restart system services such as containerd.
+- Loading and unloading kernel modules.
+
+Only the Kubernetes cluster administrator needs to access or manage the Operator namespace.
+As a best practice, establish proper security policies and prevent any other users from accessing the Operator namespace.
+
+
+CVEs
+=================================================
+
+The following is a list of known CVEs in the GPU Operator or its operands.
+To view any published security bulletins for NVIDIA products published security bulletins for NVIDIA products, refer to the NVIDIA product security page at https://www.nvidia.com/en-us/security/.
+
+.. list-table:: CVEs
+   :widths: 20 45 35
+   :header-rows: 1
+
+   * - CVE ID
+     - Affected Components
+     - Fixed Version
+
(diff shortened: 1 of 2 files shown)


gpu-operator/security.rst
```diff
@@ -0,0 +1,103 @@
...
+- Access to the host file system and hardware devices, such as NVIDIA GPUs.
+- Restart system services such as containerd.
+- Permit users to list all GPU clients using the ``nvidia-smi`` utility.
+
+Only the Kubernetes cluster administrator needs to access or manage the Operator namespace.
+As a best practice, establish proper security policies and prevent any other users from accessing the Operator namespace.
+
+
+
+CVEs
```
a-mccarthy, line 28: Do we think a table like this is useful? Are their other fields we should maintain in our docs (all the CVE info is also in the linked bulletins)

There are a few more of these CVEs that relate to vulnerabilities in images used in older version. For completeness, we should probably include them as well, but i wanted to make sure that the table was valuable before i went to the trouble of adding all the rest.

gpu-operator/security.rst
```diff
@@ -0,0 +1,103 @@
...
+
+       NVIDIA GPU Operator, all versions up to and including 24.6.2
+     - NVIDIA Container Toolkit 1.17.0
+
+       NVIDIA GPU Operator 24.9.0
+
+   * - `NVIDIA CVE-2024-0132 <https://nvidia.custhelp.com/app/answers/detail/a_id/5582>`_
+     - NVIDIA Container Toolkit, all versions up to and including 1.16.1
+
+       NVIDIA GPU Operator, all versions up to and including 24.6.2
```
cdesiniotis, line 85:
```suggestion
       NVIDIA GPU Operator, all versions up to and including 24.6.1

```

gpu-operator/security.rst
cdesiniotis: I think it is valuable. I don't think we should add any more information to this table though. The security bulletin should be the source of truth and readers should be redirected there if they want more information about any particular CVE.

cdesiniotis (commented):
@a-mccarthy Thanks for this PR! I am good with these changes, but I would like to get another opinion on the table contents.

cdesiniotis:
ping @tariq1890 @francisguillier

gpu-operator/security.rst
```diff
@@ -0,0 +1,103 @@
...
+
+- ``privileged: true``
+- ``hostPID: true``
+- ``hostIPC: true``
+
+The elevated privileges are required for the following reasons:
+
+- Access to the host file system and hardware devices, such as NVIDIA GPUs.
+- Restart system services such as containerd.
+- Permit users to list all GPU clients using the ``nvidia-smi`` utility.
```
tariq1890, line 21: I don't think this is a valid reason:
```
Permit users to list all GPU clients using the ``nvidia-smi`` utility.
```

Let's add `Loading and unloading kernel modules` to the list

gpu-operator/security.rst
a-mccarthy: updated

tariq1890 (approved):

PR #185, opened 2025-06-02
Add tooling to bump container-toolkit version

elezar:


container-toolkit/release-notes.md
```diff
@@ -6,9 +6,36 @@
...
+### Fixes and Features
+
+- Updated the ordering of Mounts in CDI to have a deterministic output. This makes testing more consistent.
+- Added NVIDIA_CTK_DEBUG envvar to hooks.
+
+#### Enhancements to libnvidia-container
+
+- Fixed a bug in setting the default for the `--cuda-compat-mode` flag. This caused failures in use cases invoking the `nvidia-container-cli` directly or when an older `nvidia-container-toolkit` version was used with a newer `nvidia-container-cli`.
+- Added additional logging to the `nvidia-container-cli`.
+- Fixed variable initialisation when updating the ldcache. This caused failures on Arch linux or other platforms where the `nvidia-container-cli` was built from source.
```
a-mccarthy, line 24:
```suggestion
- Fixed variable initialization when updating the ldcache. This caused failures on Arch linux or other platforms where the `nvidia-container-cli` was built from source.
```

nit

container-toolkit/release-notes.md
```diff
@@ -6,9 +6,36 @@
...
 
-______________________________________________________________________
+## NVIDIA Container Toolkit 1.17.8
+
+This release of the NVIDIA Container Toolkit `v1.17.8` is a bugfix release.
+
+### Fixes and Features
+
+- Updated the ordering of Mounts in CDI to have a deterministic output. This makes testing more consistent.
+- Added NVIDIA_CTK_DEBUG envvar to hooks.
```
a-mccarthy, line 18:
```suggestion
- Added NVIDIA_CTK_DEBUG environment variable to hooks.
```

i think it makes sense to spell this out for clarity (unless envvar is something else)

container-toolkit/install-guide.md
```diff
@@ -56,7 +56,12 @@ Alternatively, you can install the driver by [downloading](https://www.nvidia.co
 1. Install the NVIDIA Container Toolkit packages:
 
    ```console
-   $ sudo apt-get install -y nvidia-container-toolkit
+   $ NVIDIA_CONTAINER_TOOLKIT_VERSION=1.17.7-1 \
```
a-mccarthy, line 59: should this be 1.17.7-1 for every release? Should we make it a variable so that the version numbers updates?

container-toolkit/release-notes.md
```diff
@@ -6,9 +6,36 @@
...
+- `nvidia-container-toolkit-base 1.17.8`
+- `libnvidia-container-tools 1.17.8`
+- `libnvidia-container1 1.17.8`
+
+The following `container-toolkit` containers are included:
+
+- `nvcr.io/nvidia/k8s/container-toolkit:v1.17.8-ubi8`
+- `nvcr.io/nvidia/k8s/container-toolkit:v1.17.8-ubuntu20.04` (also as `nvcr.io/nvidia/k8s/container-toolkit:v1.17.8`)
 
 ## NVIDIA Container Toolkit 1.17.7
```
a-mccarthy, line 40: Do you think its necessary to update the known issue section of 1.17.7 with a recommendation to upgrade to using 1.17.8 to avoid the bug? possibly unnecessary, since we already mention in the current note that the issue is fixed in the next patch release

scripts/bump-version.sh
a-mccarthy, line 1: just to confirm, this script is to allow folks to update all the version information for a component without having to manually update the files?

Some notes, all the components have a version (old format) and version1 (new format) file. So we'd have to also update the version file for each component to use the new format, right?

container-toolkit/release-notes.md
cdesiniotis: +1

container-toolkit/install-guide.md
cdesiniotis: Maybe we can update this to 1.17.8-1 in this PR and then look to parameterize it in a follow-up?

cdesiniotis (dismissed):
@a-mccarthy the overall content lgtm. Feel free to directly apply your suggestions / nits. I think we can proceed and merge this while @elezar is out.

repo.toml
```diff
@@ -42,14 +42,14 @@ sphinx_conf_py_extra = """
 
   html_domain_indices = False
   html_use_index = False
-  html_extra_path = ["versions1.json"]
+  html_extra_path = ["versions.json"]
```
mikemckiernan, line 45: There were two reasons for using `versions1.json`:

- It avoided clobbering the older format that was used before NVIDIA Sphinx Theme.  This way, if someone stumbled on an older version of the docs, the version menu would still work.
- Naming the file `versions1.json` is a bit of a convention across docs projects at NVIDIA, kinda because we've nearly all migrated to using NVIDIA Sphinx Theme.

IMO, not worth changing and kind of worth it to keep as-is for continuity across projects.

mikemckiernan (commented):
Before doing it this way, please consider using `source_subtitutions`.

container-toolkit/release-notes.md
elezar: Sure. Let's use the full text instead.

scripts/bump-version.sh
elezar: Yes, that was the intent of this script. Note that I still don't see the purpose for having both version files present in the repo at a given time -- but maybe my understanding of the document generation is incomplete.

container-toolkit/install-guide.md
elezar: This was moved to #190

PR #193, opened 2025-06-13
Correct version pinning instructions

elezar:
See https://github.com/NVIDIA/nvidia-container-toolkit/issues/1142


a-mccarthy (approved):
thanks @elezar!

PR #249, opened 2025-09-13
Update CDI support copy

chenopis:
Signed-off-by: Andrew Chen <andrewch@nvidia.com>


mikemckiernan (approved):
Makes sense.  Thanks for the fix!

PR #284, opened 2025-10-15
migrate refrences from gitlab to github

shivakunv:


README.md
```diff
@@ -8,16 +8,20 @@ The product documentation portal can be found at: https://docs.nvidia.com/datace
 ## Building the Container
 
 This step is optional if your only goal is to build the documentation.
-As an alternative to building the container, you can run `docker pull registry.gitlab.com/nvidia/cloud-native/cnt-docs:0.4.0`.
+As an alternative to building the container, you can run `docker pull ghcr.io/nvidia/cloud-native-docs:4`.
 
 Refer to <https://gitlab.com/nvidia/cloud-native/cnt-docs/container_registry> to find the most recent tag.
 
 If you change the `Dockerfile`, update `CONTAINER_RELEASE_IMAGE` in the `gitlab-ci.yml` file to the new tag and build the container.
```
shivakunv: Please check if we can remove this ?

a-mccarthy: I think this is still valid

README.md
```diff
@@ -8,16 +8,20 @@ The product documentation portal can be found at: https://docs.nvidia.com/datace
 ## Building the Container
 
 This step is optional if your only goal is to build the documentation.
-As an alternative to building the container, you can run `docker pull registry.gitlab.com/nvidia/cloud-native/cnt-docs:0.4.0`.
+As an alternative to building the container, you can run `docker pull ghcr.io/nvidia/cloud-native-docs:4`.
```
a-mccarthy: @chenopis since you've been digging around in the CI recently, can you review this section and give your feedback? Is the tag still right? should it be set to github or gitlab.

We build the docs from a gitlab mirror, so i think its ok to leave as gitlab here.

PR #347, opened 2026-02-11
Dra updates

a-mccarthy:


gpu-operator/dra-intro-install.rst
```diff
@@ -2,116 +2,336 @@
...
-=================================
+      Using the NVIDIA DRA Driver for GPUs allocation with the GPU Operator requires the following:
 
-NVIDIA's DRA Driver for GPUs is comprised of two subsystems that are largely independent of each other: one manages GPUs, and the other one manages ComputeDomains.
+      * Kubernetes v1.34.2 or newer.
 
-Below, you can find instructions for how to install both parts or just one of them.
-Additionally, NVIDIA has prepared two separate documentation chapters, providing more in-depth information for each of the two subsystems:
+        .. note::
+           If you plan to use extended resources with the DRA Driver you must also have the required feature gates installed.
```
shivamerla:
```suggestion
           If you plan to use traditional extended resource requests such as `nvidia.com/gpu` with the DRA driver, you must enable the [DRAExtendedResource](https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#extended-resource) feature gate. This feature allows the scheduler to automatically translate extended resource requests into ResourceClaims, which are then allocated by the DRA driver.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -2,116 +2,336 @@
...
+
+            helm upgrade --install gpu-operator nvidia/gpu-operator \
+              --version=${version} \
+              --create-namespace \
+              --namespace gpu-operator \
+              --set devicePlugin.enabled=false \
+              --set driver.manager.env[0].name=NODE_LABEL_FOR_GPU_POD_EVICTION \
+              --set driver.manager.env[0].value="nvidia.com/dra-kubelet-plugin"
+
+         Make sure that the value of ``driver.manager.env`` matches the node selector label for nodes that will support GPU allocation through the DRA Driver.
```
shivamerla:
```suggestion
         Make sure that the value of ``driver.manager.env`` matches the node selector label that was used when installing the DRA driver helm chart.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -2,116 +2,336 @@
...
+         .. tab-set::
+
+            .. tab-item:: Default values.yaml file
+
+             Specifies the node selector label for nodes that will support GPU allocation through the DRA Driver.
+
+             .. code-block:: yaml
+
+                nodeSelector:
+                  nvidia.com/dra-kubelet-plugin: "true"
```
shivamerla: This is not required, we need the nodeSelector label only for the kubelet plugin below.

gpu-operator/dra-intro-install.rst
```diff
@@ -2,116 +2,336 @@
...
+                # Specify the driver root on the host in nvidiaDriverRoot.
+                # "/home/kubernetes/bin/nvidia" is the default driver root on GKE.
+                nvidiaDriverRoot: "/home/kubernetes/bin/nvidia"
 
-   Example for *Operator-provided* GPU driver:
+                controller:
+                  priorityClassName: ""
+                  affinity: null
+                nodeSelector:
+                  nvidia.com/dra-kubelet-plugin: "true"
```
shivamerla: same here, this is not required, by default this is set to run the controller on the control-plane nodes which should be left as-is.

gpu-operator/dra-intro-install.rst
a-mccarthy: removed!

gpu-operator/dra-intro-install.rst
a-mccarthy: removed

PR #394, opened 2026-05-11
Minor nits in coco docs

a-mccarthy:
* adds come clarification text to the run a workload sample
* re-adds image pull timeout to the prereqs
~* improves attestation workflow to read clearer.~

### DIFF
--- a/confidential-containers/attestation.rst
+++ b/confidential-containers/attestation.rst
@@ -69,8 +69,7 @@ Provision Trustee
 Trustee is an open-source framework used in Confidential Containers to verify attestation evidence and conditionally release secrets.
 For a full overview of attestation with Trustee, refer to the upstream `Trustee documentation <https://confidentialcontainers.org/docs/attestation/>`_.
 
-To provision a Trustee instance, follow the upstream `Install Trustee in Docker <https://confidentialcontainers.org/docs/attestation/installation/docker/>`_ guide.
-This is the recommended install method.
+To provision a Trustee instance, follow the recommended upstream `Install Trustee in Docker <https://confidentialcontainers.org/docs/attestation/installation/docker/>`_ guide.
 
 .. note::
 
@@ -86,20 +85,21 @@ After you complete installation, Trustee is configured to use the NVIDIA Remote 
 Configure Workloads for Attestation
 ====================================
 
-To enable attestation for your workloads, point them to the Trustee network endpoint, sometimes referred to as the Key Broker Service (KBS) endpoint, by adding the following annotation to your workload pod spec:
+To enable attestation for your workloads, point them to the Trustee network endpoint, also called the Key Broker Service (KBS) endpoint, by adding the following annotation to your workload pod spec:
 
 .. code-block:: yaml
 
    io.katacontainers.config.hypervisor.kernel_params: "agent.aa_kbc_params=cc_kbc::http://<kbs-ip>:<kbs-port>"
 
-Replace ``<kbs-ip>`` with the IP address or hostname at which your Trustee instance is reachable from the worker nodes, and ``<kbs-port>`` with the port (default: ``8080``).
+Replace ``<kbs-ip>`` with the IP address or hostname at which your Trustee instance is reachable from the worker nodes.
+Replace ``<kbs-port>`` with the port that Trustee listens on (default: ``8080``).
 
 Refer to the upstream `Setup Confidential Containers <https://confidentialcontainers.org/docs/attestation/coco-setup/>`_ documentation for more information on configuring workloads for attestation.
 
 .. _customize-attestation:
 
-Customize Attestation Workflows
-===============================
+Optional: Customize Attestation Workflows
+=========================================
 
 After Trustee is provisioned and workloads are configured, you can customize attestation workflows to enforce your desired security policies.
 This can include configuring the following:
@@ -108,7 +108,7 @@ This can include configuring the following:
   Refer to the upstream documentation on `using the KBS Client Tool <https://confidentialcontainers.org/docs/attestation/client-tool/>`_.
 * Configure resources: Create resources, or secrets, that your workloads need. 
   Refer to the upstream `Confidential Containers resources <https://confidentialcontainers.org/docs/attestation/resources/>`_ documentation for more information on the resources.
-* Configure policies: Confidential Containers uses different policy types to secure workload at different layers.
+* Configure policies: Confidential Containers uses different policies to secure workload at different layers.
   Refer to the upstream `Confidential Containers policy <https://confidentialcontainers.org/docs/attestation/policies/>`_ documentation for more information on the policy types and configuring policies.
  
 Refer to the upstream `Confidential Containers Features <https://confidentialcontainers.org/docs/features>`_ documentation for a full list of attestation features and how to configure them.
@@ -122,6 +122,5 @@ Use the Trustee log to diagnose the attestation process.
 Next Steps
 ==========
 
-* Refer to the :doc:`deployment guide <confidential-containers-deploy>` for Confidential Containers setup instructions.
-* Refer to the upstream `Confidential Containers Features <https://confidentialcontainers.org/docs/features>`_ documentation for a complete list of attestation-dependent features.
-* Refer to the `NVIDIA Confidential Computing documentation <https://docs.nvidia.com/confidential-computing>`_ for additional information.
+* Refer to the upstream `Confidential Containers Features <https://confidentialcontainers.org/docs/features>`_ for complete documentation on attestation features.
+* If you haven't already, refer to the :doc:`Confidential Containers deployment guide <confidential-containers-deploy>` to configure your environment for confidential workloads.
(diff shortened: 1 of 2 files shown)


confidential-containers/attestation.rst
```diff
@@ -35,8 +36,14 @@ Attestation is required for any feature that depends on secrets, including:
...
 The evidence is sent to a remote verifier, Trustee, which evaluates the evidence against configured policies and conditionally releases the secret.
+Trustee is typically deployed in a separate trusted environment that is reachable from your worker nodes over the network.
+
+.. note::
 
-For background on how attestation fits into the Confidential Containers architecture, refer to the :doc:`NVIDIA Confidential Containers Reference Architecture overview <overview>`.
+  This page is an educational overview of attestation with Confidential Containers, not a complete configuration guide.
+  The attestation workflow is fully documented in the upstream `Confidential Containers documentation <https://confidentialcontainers.org/docs/attestation/>`_, which is the source of truth for setup and configuration details.
+
+  Attestation is not required to deploy Confidential Containers; it is needed only for features that rely on secret release, such as those listed above.
```
fitzthum: May want to massage this a little bit. It's true that you can start workloads without attestation, and the features above are the ones that require attestation, but key detail is that your workload isn't really secure without using attestation (via one of these features, usually). I'm not sure we really want to get into that here, though.

confidential-containers/attestation.rst
```diff
@@ -86,32 +92,34 @@ After you complete installation, Trustee is configured to use the NVIDIA Remote
...
 
 .. _customize-attestation:
 
-Customize Attestation Workflows
-===============================
+Optional: Customize Attestation Workflows
+=========================================
+
+Confidential Containers enables sensible default attestation policies for NVIDIA Confidential Computing GPUs.
+In most cases, the default policy is appropriate and you only need to provide reference values.
```
fitzthum: There is a lot packed into these two sentences.

For NVIDIA devices, you don't have to do anything regarding attestation policy or reference values. Basically, NRAS handles all of this for you.

Your guest will also have a CPU, though. Attesting the CPU does require setting reference values. You can still use the default policy, but you will need to set some reference values per the link below.

Fine to keep it as is, which is correct, and avoids some subtleties, but setting reference values for the CPU attestation is a key step for secure workloads.

confidential-containers/attestation.rst
```diff
@@ -86,32 +92,34 @@ After you complete installation, Trustee is configured to use the NVIDIA Remote
...
-* Configure resources: Create resources, or secrets, that your workloads need. 
-  Refer to the upstream `Confidential Containers resources <https://confidentialcontainers.org/docs/attestation/resources/>`_ documentation for more information on the resources.
-* Configure policies: Confidential Containers uses different policy types to secure workload at different layers.
-  Refer to the upstream `Confidential Containers policy <https://confidentialcontainers.org/docs/attestation/policies/>`_ documentation for more information on the policy types and configuring policies.
- 
-Refer to the upstream `Confidential Containers Features <https://confidentialcontainers.org/docs/features>`_ documentation for a full list of attestation features and how to configure them.
+* `Resources <https://confidentialcontainers.org/docs/attestation/resources/>`_: Create the resources, such as secrets, that your workloads need.
+* `Policies <https://confidentialcontainers.org/docs/attestation/policies/>`_: Configure the policy types that secure workloads at different layers.
```
fitzthum: nit: `policy types` -> `policies`

confidential-containers/confidential-containers-deploy.rst
```diff
@@ -180,6 +180,35 @@ Kubernetes Cluster
 
      $ sudo systemctl restart kubelet
 
+.. _configure-image-pull-timeouts:
+
+* Configure image pull timeouts. The guest-pull mechanism pulls images inside the confidential VM, which means large images can take longer to download and delay container start.
```
fitzthum: Technically speaking, pulling images inside the guest doesn't necessarily take longer than pulling on the host. The issue is more that guest pull invalidates any kind of caching that would normally happen. Basically you have to pull the image every time you run a pod.

confidential-containers/confidential-containers-deploy.rst
```diff
@@ -461,7 +490,7 @@ For further configuration settings, refer to the following sections:
 Run a Sample Workload
 =====================
 
-A pod manifest for a confidential container GPU workload requires that you specify the ``kata-qemu-nvidia-gpu-snp`` runtime class for SEV-SNP or ``kata-qemu-nvidia-gpu-tdx`` for TDX.
+A pod manifest for a confidential container GPU workload requires that you specify the ``kata-qemu-nvidia-gpu-snp`` runtime class for AMD based systems or ``kata-qemu-nvidia-gpu-tdx`` for Intel based systems.
```
fitzthum: nit: hyphen in `AMD-based`? (and intel too)

fitzthum (commented):
Nice. 

Most of my comments probably go beyond the level of detail needed in the doc, but maybe they can help massage things a little bit.

confidential-containers/attestation.rst
manuelh-dev: +1 on this. [We can phrase it that attestation is not required to get a pod going for evaluation purposes but trust in the deployment can only be established via attestation **and** via other CoCo mechanisms such as image signature verification, the use of a proper agent security policy - all this is documented in the upstream repo as pointed out]

confidential-containers/attestation.rst
```diff
@@ -86,32 +92,34 @@ After you complete installation, Trustee is configured to use the NVIDIA Remote
 Configure Workloads for Attestation
 ====================================
 
-To enable attestation for your workloads, point them to the Trustee network endpoint, sometimes referred to as the Key Broker Service (KBS) endpoint, by adding the following annotation to your workload pod spec:
+To enable attestation for your workloads, point them to the Trustee network endpoint, also called the Key Broker Service (KBS) endpoint, by adding the following annotation to your workload pod spec:
 
 .. code-block:: yaml
 
    io.katacontainers.config.hypervisor.kernel_params: "agent.aa_kbc_params=cc_kbc::http://<kbs-ip>:<kbs-port>"
```
manuelh-dev, line 92: @fitzthum do we actually want to propose this? another method to provide this information is via init-data. should we not rather also abstract from this?

confidential-containers/attestation.rst
```diff
@@ -86,32 +92,34 @@ After you complete installation, Trustee is configured to use the NVIDIA Remote
...
-Customize Attestation Workflows
-===============================
+Optional: Customize Attestation Workflows
+=========================================
+
+Confidential Containers enables sensible default attestation policies for NVIDIA Confidential Computing GPUs.
+In most cases, the default policy is appropriate and you only need to provide reference values.
+For more information, refer to the upstream `Confidential Containers reference values <https://confidentialcontainers.org/docs/attestation/reference-values/>`_ documentation.
+
+You can use the Key Broker Service (KBS) Client Tool to configure Trustee reference values and secrets.
```
manuelh-dev: @fitzthum please comment here if you would propose adding that this tool would usually also be run in a secure environment and the tool would communicate with KBS in a secure way. If you want to omit this here, please resolve the comment.

confidential-containers/confidential-containers-deploy.rst
```diff
@@ -162,7 +162,7 @@ Kubernetes Cluster
   * ``RuntimeClassInImageCriApi``: Alpha since Kubernetes v1.29 and is not enabled by default.
     This feature gate is required to support pod deployments that use multiple snapshotters side-by-side.
 
-  Add both feature gates to your Kubelet configuration (typically ``/var/lib/kubelet/config.yaml``):
+  Add both feature gates to your Kubelet configuration (typically ``sudo vi /var/lib/kubelet/config.yaml``):
```
manuelh-dev: Not so sure. Can we just say 'typically adjust the file ...'. `sudo vi` is a bit specific and a bit loose. [There may be users which literally wouldn't be able to exit vim anymore]

confidential-containers/confidential-containers-deploy.rst
```diff
@@ -180,6 +180,35 @@ Kubernetes Cluster
...
+     kind: KubeletConfiguration
+     runtimeRequestTimeout: 20m
+
+  Restart the kubelet service to apply the change:
+
+  .. code-block:: console
+
+     $ sudo systemctl restart kubelet
+
+  Optionally, you can configure additional timeouts for the NVIDIA Shim and Kata Agent Policy.
```
manuelh-dev: Instead of NVIDIA shim -> Kata Shim.

Please do not use Kata Agent Policy here. This timeout is unrelated to the kata agent policy feature. We can just say 'kata agent' here.

confidential-containers/attestation.rst
fitzthum: I am fine with either. This way is simpler; maybe better in the short term. Init-data is an important concept, though. We should cover it at some point.

confidential-containers/attestation.rst
fitzthum: Yeah, it might be worth saying that the configuration endpoints of the KBS are privileged and the KBS Client will use a secure channel to access them.

confidential-containers/attestation.rst
a-mccarthy: i opted for keeping this as the simple version, until we figured out how we wanted to talk about this in the end to end workflow.

confidential-containers/confidential-containers-deploy.rst
a-mccarthy: i updated the wording a bit here.

manuelh-dev (approved):

PR #404, opened 2026-05-21
turn attestation into quickstart

a-mccarthy:


confidential-containers/attestation.rst
```diff
@@ -17,110 +17,216 @@
...
+
+* Trusted Execution Environment (TEE): A hardware-isolated environment, such as AMD SEV-SNP, Intel TDX, or an NVIDIA Confidential Computing GPU, that protects code and data in use.
+* Remote attestation: The process of cryptographically proving to a remote party that a TEE is running the expected, untampered software stack before that party releases secrets to it.
+* Trustee: The remote verifier in the Confidential Containers attestation flow. Trustee is composed of three cooperating services:
+
+  * Key Broker Service (KBS): The HTTP endpoint that clients and confidential guests talk to. KBS orchestrates the attestation exchange and conditionally releases resources when policy allows.
+  * Attestation Service (AS): Verifies hardware evidence presented by a guest TEE against reference values.
+  * Reference Value Provider Service (RVPS): Holds the known-good reference values that the Attestation Service compares evidence against.
+* KBS resource: A secret, for example, a key, credential, or token,  that Trustee releases to a guest when attestation succeeds. Resources are addressed by a three-part path: ``<repository>/<type>/<tag>``.
```
fitzthum: Due to some technical strangeness, it might be better to say that "most reosurces are addressed by a three-part path`.

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,216 @@
...
+* The ``kbs-client`` command-line tool installed and able to reach your Trustee instance.
+* A sample resource request that exercises the end-to-end request path.
+
+You'll know you're done when ``kbs-client`` can send a request to KBS and receive a response from the Trustee policy engine, even if that response is a policy denial.
+A denial in this quickstart is the expected, successful outcome: it confirms that the client reached KBS, the Attestation Service evaluated the request, and policy was applied.
+
+.. note::
+
+   This quickstart is for development and evaluation only. Do not use the Trustee instance you stand up here in production.
+   It does not deploy a Trusted Execution Environment (TEE), does not produce real hardware attestation evidence, and does not release any secrets to a workload. It only validates that the Trustee components are running and reachable.
```
fitzthum: nit: `It` -> `This guide`. As is, this seems to refer to Trustee, which wouldn't setup a TEE in any case.

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,216 @@
...
+
+.. code-block:: output
+
+   NAME              IMAGE                     STATUS    PORTS
+   trustee-as-1      .../attestation-service   Up        -
+   trustee-kbs-1     .../kbs                   Up        0.0.0.0:8080->8080/tcp
+   trustee-rvps-1    .../rvps                  Up        -
+
+Confirm KBS is listening on port ``8080`` by sending a POST to the attestation endpoint.
+The request body is intentionally empty. You are checking reachability, not submitting evidence.
```
fitzthum, line 156: Might be worth mentioning that it isn't recommended to curl the endpoints directly unless you are checking connectivity. Otherwise, use the kbs-client to do stuff. (Arguably we could skip the curl step entirely idk)

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,216 @@
...
+.. code-block:: console
+
+   $ ./kbs-client --help
+
+.. note::
 
-Optional: Customize Attestation Workflows
-=========================================
+   The ORAS-distributed ``kbs-client`` does not support requesting resources from inside a TEE enclave.
+   For in-enclave use, refer to the upstream `Client Tool documentation <https://confidentialcontainers.org/docs/attestation/client-tool/>`_.
```
fitzthum, line 213: Maybe mention that this kbs-client binary is mainly used for configuring Trustee as an admin.

The get-resource endpoint can be used (including inside a TEE), but it will always invoke the sample attester, which generates fake evidence for testing.

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,216 @@
...
+.. code-block:: output
+
+   WARN attester: No TEE platform detected. Sample Attester will be used.
+   WARN kbs_protocol::client::rcar_client: Authenticating with KBS failed. Perform a new RCAR handshake: ErrorInformation {
+       error_type: "https://github.com/confidential-containers/kbs/errors/PolicyDeny",
+       detail: "Access denied by policy",
+   }
+   Error: request unauthorized
+
+The first warning confirms ``kbs-client`` detected that it is not running in a TEE and fell back to the sample attester.
```
fitzthum: In this case, only the sample attester is supported, so you would get the same thing if you were in a TEE. Maybe no need for that level of detail, tho.

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,216 @@
...
+The ``PolicyDeny`` warning and the final ``request unauthorized`` error confirm that KBS received the request, the Attestation Service evaluated it, and policy rejected it.
```
fitzthum, line 242: Maybe mention that by default the trustee policy will reject requests that do not come from inside a TEE.

fitzthum (commented):
Great stuff. A few nits, but this seems mostly spot on.

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,231 @@
...
 Attestation
 ***********
 
-This page provides an overview of how to configure remote attestation for Confidential Container workloads.
-Attestation cryptographically verifies the guest Trusted Execution Environment (TEE) for the CPU and GPU before secrets are released to a workload.
-
-Attestation is required for any feature that depends on secrets, including:
+In Confidential Containers, a Trusted Execution Environment (TEE) isolates a workload from the host.
+Attestation is a process that cryptographically proves the state of the guest TEE, including both the CPU and the GPU, to a remote verifier before any secret or sensitive resource is released to the workload.
+oAttestation is required for any feature that depends on secrets, including:
```
manuelh-dev: oAttestation => Attestation.

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,231 @@
...
+For a deeper explanation of attestation, Trustee, and the full set of features, refer to the upstream `Attestation <https://confidentialcontainers.org/docs/attestation/>`_ and `Features <https://confidentialcontainers.org/docs/features>`_ sections of the Confidential Containers documentation.
+
+This quickstart runs on a standalone Linux host and does not require a Kubernetes cluster or the Confidential Containers runtime to complete.
+In a real deployment, attestation builds on the runtime setup described in the :doc:`Confidential Containers deployment guide <confidential-containers-deploy>`. Confidential workloads use Trustee to cryptographically verify their TEE before they receive secrets, encrypted container images, authenticated registries, or other sensitive resources.
+
+.. note::
+
+   This quickstart is for development and evaluation only. Do not use the Trustee instance you stand up here in production.
+   It does not deploy a Trusted Execution Environment (TEE), does not produce real hardware attestation evidence, and does not release any secrets to a workload. It only validates that the Trustee components are running and reachable.
```
manuelh-dev: nit: did we  intentionally re-introduce TEE here (already done above)

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,231 @@
...
+Clone the upstream Trustee repository.
+The repository ships with a ``docker-compose.yml`` that wires KBS, the Attestation Service, and the Reference Value Provider Service together.
 
-#. :ref:`Provision Trustee <provision-trustee>`, the remote verifier and key broker, in a trusted environment.
-#. :ref:`Configure your workloads <configure-workloads-trustee>` to point to the Trustee network endpoint.
-#. Optionally, :ref:`customize attestation workflows <customize-attestation>` for your use cases.
+.. code-block:: console
 
-After configuration, the Confidential Containers runtime automatically runs the attestation flow when a workload requires it.
+   $ git clone https://github.com/confidential-containers/trustee.git && cd trustee
```
manuelh-dev, line 111: @fitzthum fine with me but my general sentiment for such documentation would  be not having to clone repositories. Is there no way to use a release of trustee instead?

confidential-containers/confidential-containers-deploy.rst
```diff
@@ -28,7 +28,8 @@ These are key pieces of the NVIDIA Confidential Containers Reference Architectur
 
 Before you begin, refer to the :doc:`Confidential Containers Reference Architecture <overview>` for details on the reference architecture and the :doc:`Supported Platforms <supported-platforms>` page for the supported platforms.
 
-This guide assumes you are familiar with the NVIDIA GPU Operator, Kata Containers, and Kubernetes cluster administration.
+This guide is for Kubernetes cluster administrators with host access to worker nodes (for BIOS and kernel configuration) and cluster-admin access to ``kubectl``.
```
manuelh-dev: nit: cluster admin access *using* kubectl?

manuelh-dev (approved):
LGTM, a few minor comment, plus agreeing with already placed comments.

confidential-containers/attestation.rst
a-mccarthy: I adedd in a note about this.

confidential-containers/attestation.rst
a-mccarthy: i removed part of the output here, to just focus on the kbs warning, and not get into the weeds about the sample attester

confidential-containers/attestation.rst
a-mccarthy: Is there a link to the default trustee policy that we should add? We mention that this fails b/c its not coming from inside a TEE before the output

confidential-containers/attestation.rst
a-mccarthy: Yes, b/c its a key concepts. Optimizing for time, its not exactly a smooth intro. If we really want to commit to explaining terms, than we'd need a full glossary. or create one for the coco site.

confidential-containers/confidential-containers-deploy.rst
a-mccarthy: updated.

confidential-containers/attestation.rst
a-mccarthy: Are there any docs around the get-resource endpoint?

confidential-containers/attestation.rst
fitzthum: The docker compose deployment relies on having the repo. It uses pre-built artifacts, though. There are some issues with version slew here. We can resolve those in the future.

confidential-containers/attestation.rst
fitzthum: Default policy is [here](https://github.com/confidential-containers/trustee/blob/main/kbs/sample_policies/default.rego), but probably not tol link straight to the source.

confidential-containers/attestation.rst
fitzthum: Not much beyond the help for the command itself.

fitzthum (approved):
LGTM

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,238 @@
...
 Attestation is required for any feature that depends on secrets, including:
 
 * Pulling encrypted container images
 * Accessing authenticated container registries
 * Using sealed secrets
 * Requesting secrets directly from workloads
 
-When a workload requires a secret, such as a key to decrypt a container image or model, guest components collect hardware evidence from the active CPU and GPU enclaves.
-The evidence is sent to a remote verifier, Trustee, which evaluates the evidence against configured policies and conditionally releases the secret.
+When a workload requires a secret, such as a key to decrypt a container image or model, guest components collect hardware evidence from the active CPU and GPU enclaves. The evidence is sent to Trustee, the remote verifier in Confidential Containers deployments. Trustee evaluates the evidence against known-good reference values and configured policies, and conditionally releases the requested resource.
```
manuelh-dev: @fitzthum - can we avoid mentioning trustee here yet - keep it generic until key concepts?

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,238 @@
...
+* KBS resource: A secret, for example, a key, credential, or token, that Trustee releases to a guest when attestation succeeds. Most resources are addressed by a three-part path: ``<repository>/<type>/<tag>``.
+* Policy: The rule set that Trustee evaluates against verified evidence to decide whether to release a resource. By default, Trustee denies resource requests from clients that have not presented valid TEE evidence.
+
+Quickstart
+==========
+
+This page walks you through standing up a development Trustee instance with Docker Compose, installing the Key Broker Service (KBS) client tool, and sending a sample resource request to confirm the system is reachable.
+The goal is to give you a working attestation backend and a client you can use to interact with it before you wire it into a Confidential Containers workload.
+
+This page is for new users who want to try out attestation on a single Linux host.
```
manuelh-dev: @fitzthum - instead of this, can we say that this is for new users who want to try setting up trustee in a basic way - instead of exercising attestation - we don't really exercise attestation here right?

confidential-containers/attestation.rst
```diff
@@ -17,110 +17,238 @@
...
+For a deeper explanation of attestation, Trustee, and the full set of features, refer to the upstream `Attestation <https://confidentialcontainers.org/docs/attestation/>`_ and `Features <https://confidentialcontainers.org/docs/features>`_ sections of the Confidential Containers documentation.
```
manuelh-dev: @a-mccarthy @fitzthum - maybe here's a good place to re-emphasize that the full end-to-end setup is out of scope for the NVIDIA documentation here.

confidential-containers/attestation.rst
fitzthum: Fine either way imo

confidential-containers/attestation.rst
fitzthum: Yeah we should keep the part about how this is just for trying things out on a single node, but maybe we can say it's the first steps towards attestation. not sure the best way to phrase

confidential-containers/attestation.rst
fitzthum: Yeah, I think one thing to emphasize is that the upstream guide does not require any nvidia specific configurations.

Issue #391, opened 2026-05-06
Review air-gapped page


a-mccarthy:
This page has some old examples. we should review it and make sure the examples and details are still valid.

PR #409, opened 2026-05-26
docs: refresh air-gapped values example

omribz156:
Summary:
- Replace the old `v1.7.0` raw `values.yaml` fetch with the Helm chart values flow used elsewhere in the GPU Operator docs.
- Remove fixed GPU Operator v1.9-era component tags from the air-gapped sample and show a version-neutral local registry override pattern.
- Add guidance to keep image and version values from the selected chart's `values.yaml`.

Verification:
- `rg -n "v1\.7\.0|v1\.9\.0|11\.4\.2-base-ubi8|v0\.10\.0-ubi8|2\.3\.1-2\.6\.0|helm show values|Sample ``values.yaml``|<driver-version>" gpu-operator/install-gpu-operator-air-gapped.rst`
- `git diff --check`

Notes:
- I could not run `pre-commit` or `helm` locally because those CLIs are not installed in this environment.
- The `helm show values` flow is already used in the current GPU Operator docs (`getting-started.rst` and `upgrade.rst`).

Addresses part of #391.

This was implemented with Codex assistance, with the final diff manually reviewed and kept to the docs page scope.

### DIFF
--- a/gpu-operator/install-gpu-operator-air-gapped.rst
+++ b/gpu-operator/install-gpu-operator-air-gapped.rst
@@ -92,11 +92,16 @@ Before proceeding to the next sections, get the ``values.yaml`` file used for GP
 
 .. code-block:: console
 
-  $ curl -sO https://raw.githubusercontent.com/NVIDIA/gpu-operator/v1.7.0/deployments/gpu-operator/values.yaml
+  $ helm repo add nvidia https://helm.ngc.nvidia.com/nvidia \
+      && helm repo update
+
+.. code-block:: console
+
+  $ helm show values nvidia/gpu-operator --version=${version} > values.yaml
 
 .. note::
 
-   Replace ``v1.7.0`` in the above command with the version you want to use.
+   Replace ``${version}`` in the above command with the GPU Operator chart version you want to use.
 
 
 ********************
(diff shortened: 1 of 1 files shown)


gpu-operator/install-gpu-operator-air-gapped.rst
```diff
@@ -161,77 +166,65 @@ Finally, push the images to the local registry:
...
+Update ``values.yaml`` with local registry information in the repository field for each image that you mirror.
+The image names and versions change between GPU Operator releases, so start with the ``values.yaml`` file for your chart
+version and keep the image and version values from that file.
 
 .. note::
 
    Replace <repo.example.com:port> below with your local image registry URL and port.
 
-Sample of ``values.yaml`` for GPU Operator v1.9.0:
+Sample ``values.yaml`` override pattern:
```
a-mccarthy, line 177: I think this whole sample is out of date actually, not just the version numbers. 

the helm chart values are listed here: https://github.com/NVIDIA/gpu-operator/blob/main/deployments/gpu-operator/values.yaml

a-mccarthy (commented):
@omribz156 thanks for the contribution! apologies for the delay in the review.

This page is currently using a very old sample file. We should update the sample to be more recent, as it is missing components and formatting. 

Updating the version numbers is only part of the updates needed.

gpu-operator/install-gpu-operator-air-gapped.rst
omribz156: Thanks, updated in e06fe2fe. I refreshed the override sample against the current NVIDIA/gpu-operator values.yaml instead of only changing versions: added the newer image-bearing components, fixed the validator/gfd image names, split dcgm from dcgmExporter, and normalized the YAML indentation. Verified with git diff --check, parsed the sample YAML block, and ran the source-truth helper against this doc.

omribz156: Small follow-up: the first push missed the required DCO sign-off, so I amended the commit with Signed-off-by and force-pushed with lease. The current commit is 3e8603c1 with the same doc changes.

gpu-operator/install-gpu-operator-air-gapped.rst
```diff
@@ -161,85 +166,134 @@ Finally, push the images to the local registry:
...
-      enabled: true
-      repository: <repo.example.com:port>
-      image: k8s-mig-manager
-      version: v0.2.0-ubuntu20.04
-
-    node-feature-discovery:
-      image:
-        repository: <repo.example.com:port>
-        pullPolicy: IfNotPresent
-        # tag, if defined will use the given image tag, else Chart.AppVersion will be used
```
a-mccarthy, line 240: one small nit that we seem to have lost this comment in the updates

PR #430, opened 2026-06-10
Add / to versions file

a-mccarthy:
This will fix the versions switcher to link to the correct URLs


tariq1890:
I see the same problem in the Container toolkit docs too - https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/1.18.2

Can we make the same change in `versions1.json` of the other projects?

container-toolkit/versions1.json
```diff
@@ -1,71 +1,71 @@
 [
   {
     "preferred": "true",
-    "url": "../1.19.1",
```
a-mccarthy, line 4: we seem to have a lot of container toolkit versions. anyone see a problem with trimming it to be a smaller list? maybe the most recent 4?

tariq1890: I am open to that, but let's do the trimming down in a different PR

tariq1890 (approved):

PR #447, opened 2026-07-16
Add Rafay to partner validated docs

a-mccarthy:


partner-validated/rafay.rst
```diff
@@ -0,0 +1,304 @@
...
+Preparing the cluster for the GPU Operator
+==========================================
+
+Because Rafay MKS provisions upstream Kubernetes, no MKS-specific privilege or configuration changes
+are expected to be required before installing the GPU Operator.
+
+Installing the GPU Operator as a Rafay Add-On
+=============================================
+
+Install the GPU Operator by defining it as a Rafay Add-On and referencing that Add-On from a Cluster
```
tariq1890, line 214: At first, I was going to say that we shouldn't capitalise "Add-On", but I see that it has a specific meaning in the context of Rafay. Should we add a hyperlink to this page? https://docs.rafay.co/blueprints/addons/

partner-validated/rafay.rst
```diff
@@ -0,0 +1,304 @@
...
+     .. code-block:: yaml
+
+        driver:
+          enabled: true
+          repository: nvcr.io/nvidia
+          image: driver
+          version: "580.159.04"
+
+#. Reference the GPU Operator Add-On from a **Cluster Blueprint** in the same Rafay Project. The
+   Blueprint can include other add-ons the target clusters require (CNI, storage, observability, and
```
tariq1890: Any reason we don't capitalise add-ons here?

partner-validated/rafay.rst
```diff
@@ -0,0 +1,304 @@
...
+
+Install the GPU Operator by defining it as a Rafay Add-On and referencing that Add-On from a Cluster
+Blueprint applied to your GPU-enabled MKS clusters.
+
+#. In the Rafay Console, create a new Add-On for the NVIDIA GPU Operator with the following
+   configuration:
+
+   * Source: https://helm.ngc.nvidia.com/nvidia
+
+   * Version: v26.3.1
```
tariq1890, line 222: If there is a gpu-operator version with a more recent patch release, we should be recommending that IMO. Can we ask Rafay to validate their setup with v26.3.3?

partner-validated/rafay.rst
a-mccarthy: added in a link!

tariq1890 (approved):

PR #454, opened 2026-07-26
fix(scripts): make create_archive.sh sed in-place editing portable to GNU/Linux

andrewwhitecdw:
Closing per maintainer feedback — the script is unused cruft and will be removed instead. Thanks for the review.


mikemckiernan:
Thanks @andrewwhitecdw for the fix.  However, I think we'll pass on this update in favor of removing the script.  It's cruft that hasn't been used in ages, sorry for the distraction!

PR #457, opened 2026-08-10
docs: Daemon set mounts for OL nodes

mikemckiernan:


gpu-operator/platform-support.rst
```diff
@@ -477,6 +477,21 @@ Cloud Service Providers
       - 1.32---1.36
 
 
+Starting with version 26.7.0, the Operator can render driver workloads for nodes when Node Feature Discovery
+reports ``feature.node.kubernetes.io/system-os_release.ID=ol``.
+The Operator combines ``ol`` with the major component of the ``VERSION_ID`` label to form the operating system tag.
```
tariq1890: Why use `ol` as the example here? Wouldn't `rhel` or `rocky` be better given that we publish driver containers for them? 

Note that we only added support for gpu-operator to render OracleLinux driver image tags the same as it was a community contribution and the change itself was low-risk

gpu-operator/platform-support.rst
```diff
@@ -477,6 +477,21 @@ Cloud Service Providers
...
+For example, ``VERSION_ID=9.7`` produces ``ol9``.
+For tag-based driver images, the Operator appends the operating system tag to the configured driver
+version, so a version such as ``580.126.20`` resolves to an image tag that ends in ``580.126.20-ol9``.
+The Operator also mounts custom yum repository configuration in ``/etc/yum.repos.d`` and custom certificate
+configuration in ``/etc/pki/ca-trust/extracted/pem`` on these nodes.
+
+This behavior provides operating system tag selection and manifest rendering only.
```
tariq1890: Agreed that is a useful disclaimer, but I am wondering if the need for this is prompted by the fact that we used `ol` as an example to explain the new gpu-operator behaviour

gpu-operator/platform-support.rst
mikemckiernan: I made this update after seeing the item on the project board.  I missed the other PR that addressed rhel/rocky.  I've revised and hope you can take a look again.

I'm OK with removing mention of OL if you prefer.  lmk and thanks!

mikemckiernan: @tariq1890 , revised.  PLMK if it's closer to the truth and thanks!

tariq1890: Sorry for the late response @mikemckiernan.

> I'm OK with removing mention of OL if you prefer. lmk and thanks!

I think it's preferable to not mention OL here. Instead of OL, let's just stick to either RHEL or Rocky.

PR #458, opened 2026-08-10
docs: DRA updates for 26.7.0

mikemckiernan:
Primary page for DRA: https://nvidia.github.io/cloud-native-docs/review/pr-458/gpu-operator/latest/dra-intro-install.html

### DIFF
--- a/gpu-operator/dra-intro-install.rst
+++ b/gpu-operator/dra-intro-install.rst
@@ -28,19 +28,14 @@ Before using the DRA Driver for NVIDIA GPUs, familiarize yourself with the follo
 * `Dynamic Resource Allocation <https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/>`_ in the Kubernetes documentation.
 * `DRA Driver for NVIDIA GPUs documentation <https://dra-driver-nvidia-gpu.sigs.k8s.io/docs/>`__.
 
-.. important::
-
-   GPU Operator supports the DRA Driver as a more recent alternative to the Device Plugin for Kubernetes.
-   However, some features of the DRA driver are alpha maturity and not fully supported.
-   The driver does not provide feature parity with the device plugin.
-   Refer to the following section regarding the features that are not yet implemented or fully supported before selecting the driver for production use.
-
 *********************************
 Comparison: DRA and Device Plugin
 *********************************
 
-The DRA Driver for NVIDIA GPUs and the NVIDIA Device Plugin for Kubernetes provide alternative mechanisms for allocating NVIDIA GPU resources.
-The mechanisms do not provide feature parity.
+GPU Operator supports the DRA Driver as a more recent alternative to the NVIDIA Device Plugin for Kubernetes.
+However, some features of the DRA driver are alpha maturity and not fully supported.
+The driver does not provide feature parity with the device plugin.
+
 A cluster can have either a ``GPUCluster`` resource for DRA or a ``ClusterPolicy`` resource for the Device Plugin, but not both.
 
 Use DRA for workloads that have the following requirements:
@@ -117,7 +112,7 @@ The following table groups the DRA driver capabilities by maturity:
    :widths: 34 22 18 26
 
    * - Capability
-     - Maturity
+     - Support Status
      - Default
      - Operator Control
 
@@ -432,10 +427,19 @@ During reconciliation, the STATUS column progresses from empty to `notReady` to 
 
    The ComputeDomain DeviceClasses are present only when ComputeDomain support is enabled.
 
-Additional validation procedures are available in the upstream DRA Driver documentation:
+*************************
+Running a Sample Workload
+*************************
 
-* `Run a sample ComputeDomain workload <https://dra-driver-nvidia-gpu.sigs.k8s.io/docs/install/#run-a-sample-computedomain-workload>`__.
-* `Run a sample GPU allocation workload <https://dra-driver-nvidia-gpu.sigs.k8s.io/docs/install/#run-a-sample-gpu-allocation-workload>`__.
+Refer to `Request full GPUs <https://dra-driver-nvidia-gpu.sigs.k8s.io/docs/guides/gpu-allocation/allocating-gpus/>`__
+in the DRA driver documentation for information about the following workload-related tasks:
+
+* Request any GPU
+* Request multiple GPUs in one pod
+* Share a GPU across containers in a pod
+* Select a GPU by product name
+* Select a GPU by memory size
+* Combine attribute and capacity selectors
 
 *********
 Telemetry
--- a/gpu-operator/dra-intro-install.rst
+++ b/gpu-operator/dra-intro-install.rst
@@ -33,24 +33,9 @@ Comparison: DRA and Device Plugin
 *********************************
 
 GPU Operator supports the DRA Driver as a more recent alternative to the NVIDIA Device Plugin for Kubernetes.
-However, some features of the DRA driver are alpha maturity and not fully supported.
-The driver does not provide feature parity with the device plugin.
+However, some features of the DRA driver have alpha support status.
 
 A cluster can have either a ``GPUCluster`` resource for DRA or a ``ClusterPolicy`` resource for the Device Plugin, but not both.
-
-Use DRA for workloads that have the following requirements:
-
-* Coordinate Multi-Node NVLink workloads by using ComputeDomains.
-* Use attribute-based GPU selection or allocation-specific device configuration instead of node-wide configuration.
-* Configure dynamic NVIDIA Multi-Instance GPU (MIG), CUDA Multi-Process Service (MPS),
-  CUDA time-slicing, or Virtual Function I/O (VFIO) passthrough for individual allocations.
-
-Use the Device Plugin for workloads that have the following requirements:
-
-* Request ``nvidia.com/gpu`` or MIG extended resources from existing Pod specifications and tools.
-* Schedule CUDA time-slicing or MPS replicas as independent extended resources.
-* Use the broader set of components managed through ``ClusterPolicy``, Device Plugin health reporting,
-  or Kubernetes Pod priority and preemption.
 
 Capability Comparison
 =====================
@@ -79,7 +64,9 @@ The following table compares the GPU allocation capabilities of the two mechanis
        MIG Manager configures MIG geometry before allocation.
 
    * - GPU Sharing
-     - Supports time-slicing through a shared DRA claim. Custom time-slice intervals require the Alpha ``TimeSlicingSettings`` feature gate.
+     - Alpha ``ConsumableShares`` feature gate implements Kubernetes consumable capacity for scheduler-accounted sharing among independent ``ResourceClaim`` objects for full GPUs and MIG devices.
+       The driver also supports time-slicing through a shared DRA claim.
+       Custom time-slice intervals require the Alpha ``TimeSlicingSettings`` feature gate.
        MPS requires the Alpha ``MPSSupport`` gate.
      - Advertises time-slicing replicas and experimental MPS replicas for all GPUs on a node.
 
@@ -131,7 +118,7 @@ The following table groups the DRA driver capabilities by maturity:
      - Enabled
      - ``draDriver.featureGates``
 
-   * - ``DeviceMetadata``, ``DynamicMIG``, ``MPSSupport``, ``NVMLDeviceHealthCheck``, ``PassthroughSupport``, ``TimeSlicingSettings``
+   * - ``ConsumableShares``, ``DeviceMetadata``, ``DynamicMIG``, ``MPSSupport``, ``NVMLDeviceHealthCheck``, ``PassthroughSupport``, ``TimeSlicingSettings``
      - Alpha
      - Disabled
      - ``draDriver.featureGates``
@@ -148,8 +135,13 @@ Consider the following limitations before selecting DRA driver v${dra_version}:
 * ``NVMLDeviceHealthCheck`` is alpha and disabled by default.
   The gRPC health probe reports kubelet plugin availability, and DCGM provides telemetry.
   Neither provides DRA allocation health status.
-* This release does not provide scheduler-accounted capacity sharing among independent ``ResourceClaim`` objects.
-  Workloads must share one claim instead of requesting independent replicas.
+* Consumable capacity is available through the alpha ``ConsumableShares`` feature gate, which is disabled by default.
+  When enabled, the feature provides scheduler-accounted capacity sharing among independent ``ResourceClaim`` objects for full GPUs and MIG devices.
+  Kubernetes v1.34 and v1.35 require enabling the ``DRAConsumableCapacity`` feature gate.
+  Kubernetes v1.36 and later enable the feature by default.
+  You must also set the ``CONSUMABLE_SHARES`` environment variable for the DRA GPU kubelet plugin to ``memory``, ``unlimited``, or a positive integer.
+  Set the environment variable through ``draDriver.gpus.kubeletPlugin.env``.
+  MPS cannot be used when consumable capacity is configured through ``ConsumableShares``.
 * Alpha feature gates have compatibility constraints.
   ``DynamicMIG`` conflicts with ``PassthroughSupport``, ``NVMLDeviceHealthCheck``, and ``MPSSupport``.
   ``PassthroughSupport`` conflicts with ``NVMLDeviceHealthCheck``, and ``DeviceMetadata`` requires ``PassthroughSupport``.
--- a/gpu-operator/dra-intro-install.rst
+++ b/gpu-operator/dra-intro-install.rst
@@ -20,8 +20,7 @@
 DRA Driver for NVIDIA GPUs
 ##########################
 
-Dynamic Resource Allocation (DRA) is a Kubernetes API for flexibly requesting,
-configuring, and sharing specialized devices such as GPUs.
+Dynamic Resource Allocation (DRA) is a Kubernetes API for flexibly requesting, configuring, and sharing specialized devices such as GPUs.
 This page describes how to use the GPU Operator to install and manage DRA Driver for NVIDIA GPUs v${dra_version}.
 
 Before using the DRA Driver for NVIDIA GPUs, familiarize yourself with the following documentation:
@@ -31,14 +30,16 @@ Before using the DRA Driver for NVIDIA GPUs, familiarize yourself with the follo
 
 .. important::
 
-   GPU Operator management of DRA is available as a technology preview.
-   Technology preview features are not supported in production environments and are not functionally complete.
+   GPU Operator supports the DRA Driver as a more recent alternative to the Device Plugin for Kubernetes.
+   However, some features of the DRA driver are alpha maturity and not fully supported.
+   The driver does not provide feature parity with the device plugin.
+   Refer to the following section regarding the features that are not yet implemented or fully supported before selecting the driver for production use.
 
 *********************************
 Comparison: DRA and Device Plugin
 *********************************
 
-The DRA Driver for NVIDIA GPUs and the NVIDIA Kubernetes Device Plugin provide alternative mechanisms for allocating NVIDIA GPU resources.
+The DRA Driver for NVIDIA GPUs and the NVIDIA Device Plugin for Kubernetes provide alternative mechanisms for allocating NVIDIA GPU resources.
 The mechanisms do not provide feature parity.
 A cluster can have either a ``GPUCluster`` resource for DRA or a ``ClusterPolicy`` resource for the Device Plugin, but not both.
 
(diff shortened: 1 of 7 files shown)


gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
 DRA Driver for NVIDIA GPUs
 ##########################
 
-Dynamic Resource Allocation (DRA) is a Kubernetes concept for flexibly requesting, configuring, and sharing specialized devices like GPUs.
-This page describes how to install and upgrade the DRA Driver for NVIDIA GPUs v${dra_version} with the NVIDIA GPU Operator.
+Dynamic Resource Allocation (DRA) is a Kubernetes API for flexibly requesting,
+configuring, and sharing specialized devices such as GPUs.
```
bmozaffa: Why a line break?

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
-*****************
-
-With the DRA Driver for NVIDIA GPUs, your Kubernetes workloads can allocate and consume the following two types of resources:
+.. important::
 
-* GPU allocation: for controlled sharing and dynamic reconfiguration of GPUs.
-  This functionality replaces the traditional GPU allocation method used by the NVIDIA Kubernetes Device Plugin.
-* ComputeDomains: an abstraction for secure `Multi-Node NVLink (MNNVL) <https://docs.nvidia.com/multi-node-nvlink-systems/index.html>`_ for NVIDIA GB200 and similar systems.
+   GPU Operator management of DRA is available as a technology preview.
+   Technology preview features are not supported in production environments and are not functionally complete.
```
bmozaffa: As per our decision, these two lines should be replaced with some version of this:

GPU Operator supports the DRA Driver as a modern alternative to the Device Plugin, but certain features of the DRA Driver are in alpha and not fully supported. As such, the DRA Driver does not currently provide full feature parity with the Device Plugin and it is important that users gain awareness of the features that are not yet implemented or fully supported in the DRA Driver before selecting it for their production needs.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
 
-You can use these features independently or together in the same cluster.
+*********************************
+Comparison: DRA and Device Plugin
+*********************************
 
-.. _known-issues:
+The DRA Driver for NVIDIA GPUs and the NVIDIA Kubernetes Device Plugin provide alternative mechanisms for allocating NVIDIA GPU resources.
+The mechanisms do not provide feature parity.
```
bmozaffa: I put a version of this above under the important note that replaces the previously discussed Tech Preview statement

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+A cluster can have either a ``GPUCluster`` resource for DRA or a ``ClusterPolicy`` resource for the Device Plugin, but not both.
+
+Use DRA for workloads that have the following requirements:
```
bmozaffa: Is this a good approach? Bear in mind that these are not parallel alternatives that depend on user needs, but rather DRA is on track to permanently replace the device plugin, but is still not as mature. If we wanted to use this type of comparison, we would have to focus on still missing features in DRA, and for its alpha features say that use device plugin if you requires supported production usage of the feature, and basically nothing else.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+   :header-rows: 1
+   :widths: 23 38 39
+
+   * - Capability
+     - DRA with ``GPUCluster``
+     - Device Plugin with ``ClusterPolicy``
+
+   * - API and Device Selection
+     - Uses DRA API objects for structured device selection and per-claim configuration.
+     - Uses Kubernetes extended resource names and counts.
```
bmozaffa, line 57: DRA supports this in k8s 1.36+ as backward compatibility with extended resources

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+       Configuration applies at the node or cluster level.
+
+   * - Full GPU and MIG Device Allocation
+     - Allocates full GPU devices and existing MIG devices.
+       Alpha ``DynamicMIG`` creates MIG devices for a claim.
+     - Allocates full GPU devices or preconfigured MIG devices.
```
bmozaffa, line 63: Why say "existing" before and "preconfigured" now?

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+     - Uses ``ResourceClaim`` objects and DRA scheduler integration.
+       Kubernetes does not support preemption for DRA resources.
+     - Uses Kubernetes extended resource scheduling and supports Kubernetes Pod priority and preemption.
+
+DRA Driver Feature Maturity
+===========================
+
+GPU Operator support for deploying DRA components and the maturity of capabilities in the upstream DRA driver are separate considerations.
+Upstream describes ComputeDomains as *officially supported* rather than *generally available (GA)*.
+For this comparison, the table treats the terms as equivalent.
```
bmozaffa: These 3 lines no longer apply in the same way

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+``GPUCluster`` always enables GPU allocation, can disable ComputeDomains, and passes ``draDriver.featureGates`` values to the DRA driver.
+
+The following table groups the DRA driver capabilities by maturity:
+
+.. list-table:: DRA Driver Capability Maturity
+   :header-rows: 1
+   :widths: 34 22 18 26
+
+   * - Capability
+     - Maturity
```
bmozaffa: The Maturity column here is problematic. We should not communicate the beta API as a lack of maturity of the feature. We can use the alpha API to say the feature is not GA, but for any other scenarios, the right classification is simply GA

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+.. list-table:: GPU Allocation Capability Comparison
+   :header-rows: 1
+   :widths: 23 38 39
+
+   * - Capability
+     - DRA with ``GPUCluster``
+     - Device Plugin with ``ClusterPolicy``
+
+   * - API and Device Selection
+     - Uses DRA API objects for structured device selection and per-claim configuration.
```
bmozaffa:
```suggestion
     - Uses DRA API to select GPUs by attributes (model, architecture, memory, compute capability, UUID, addressing mode, PCI/topology information) using CEL expressions.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+     - Uses Kubernetes extended resource names and counts.
+       Configuration applies at the node or cluster level.
```
bmozaffa:
```suggestion
       Node labeling or other external mechanisms required for structured selection.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+     - Reports unhealthy devices through the Kubernetes Device Plugin API.
+
+   * - Device Injection and Managed Components
+     - Requires a Container Device Interface (CDI)-compatible runtime.
+       ``GPUCluster`` manages DRA, ComputeDomains, NVIDIA Data Center GPU Manager (DCGM), DCGM Exporter, and the DRA validator.
+     - Supports multiple device injection strategies.
+       ``ClusterPolicy`` manages NVIDIA Container Toolkit, GPU Feature Discovery, MIG Manager, the sandbox device plugin, Kata Containers, KubeVirt, and NVIDIA vGPU Manager.
+
+   * - Kubernetes Scheduling
+     - Uses ``ResourceClaim`` objects and DRA scheduler integration.
```
bmozaffa:
```suggestion
     - Allocation through a ``ResourceClaim``, which can be referenced by multiple containers or Pods.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+       Kubernetes does not support preemption for DRA resources.
```
bmozaffa:
```suggestion
```

Not sure what you mean here. Not aware of a use case that cannot be fulfilled.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+     - Uses Kubernetes extended resource scheduling and supports Kubernetes Pod priority and preemption.
```
bmozaffa:
```suggestion
     - Allocation through the container's resource limits.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+       Configuration applies at the node or cluster level.
+
+   * - Full GPU and MIG Device Allocation
+     - Allocates full GPU devices and existing MIG devices.
+       Alpha ``DynamicMIG`` creates MIG devices for a claim.
+     - Allocates full GPU devices or preconfigured MIG devices.
+       MIG Manager configures MIG geometry before allocation.
+
+   * - GPU Sharing
+     - Alpha ``TimeSlicingSettings`` and ``MPSSupport`` configure sharing through DRA claims.
```
bmozaffa:
```suggestion
     - Supports time-slicing through a shared DRA claim. Custom time-slice intervals require the Alpha ``TimeSlicingSettings`` feature gate. MPS requires the Alpha ``MPSSupport`` gate.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+     - Advertises time-slicing replicas and experimental MPS replicas as independent extended resources.
```
bmozaffa:
```suggestion
     - Advertises time-slicing replicas and experimental MPS replicas for all GPUs on a node.
```

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+Refer to the `DRA driver v${dra_version} overview <https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/v${dra_version}/README.md#overview>`__
+and `feature-gate definitions <https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/v${dra_version}/pkg/featuregates/featuregates.go>`__
+for the authoritative support status, maturity, and default values.
+
+DRA Driver Limitations
+======================
+
+Consider the following limitations before selecting DRA driver v${dra_version}:
+
+* ComputeDomains are officially supported, but full GPU and MIG allocation are not yet officially supported upstream.
```
bmozaffa: This is incorrect. It's just Dynamic MIG that's still alpha.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
+  ``PassthroughSupport`` conflicts with ``NVMLDeviceHealthCheck``, and ``DeviceMetadata`` requires ``PassthroughSupport``.
+* Kubernetes does not support preemption for DRA resources.
+  Injecting DRA-allocated devices requires a preconfigured CDI-compatible container runtime.
+  The managed ``gpu.nvidia.com`` ``DeviceClass`` does not set ``spec.extendedResourceName`` to ``nvidia.com/gpu``.
+
+********
+Overview
+********
 
+The GPU Operator manages DRA components through the ``nvidia.com/v1alpha1`` ``GPUCluster`` custom resource.
```
bmozaffa:
```suggestion
The GPU Operator manages DRA components through the ``GPUCluster`` custom resource.
```

Given the decision to GA this integration, specifying the namespace is unnecessary and the alpha level API will be misleading.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
 
-.. _computedomain-prereqs:
+GPUCluster Limitations
+======================
 
-* For ComputeDomain ensure the following:
+* Use this workflow for a new installation.
+  An in-place migration from a standalone DRA driver Helm release or from ``ClusterPolicy`` to ``GPUCluster`` is not supported.
+* ``GPUCluster`` does not expose the controller affinity, priority class, toleration,
+  and kubelet-plugin node selector overrides that were used by the previous standalone DRA driver procedure for Google Kubernetes Engine.
```
bmozaffa:
```suggestion
  and kubelet-plugin node selector overrides that were used by the previous standalone DRA driver procedure for GKE (Google Kubernetes Engine).
```

The spelled out "Google Kubernetes Engine" is unfamiliar to most readers, so better to either omit it entirely, or at best put it in brackets next to the more familiar GKE

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,403 @@
...
 
-* Kubernetes v1.34.2 or later.
+For ``GPUCluster``, the GPU Operator manages the following components:
 
-  .. note::
-     If you plan to use traditional extended resource requests such as ``nvidia.com/gpu`` alongside the DRA driver, you must enable the `DRAExtendedResource <https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#extended-resource>`_ feature gate. 
-     This enabled by default in Kubernetes v1.36.0 and later. 
-     It allows the scheduler to translate extended resource requests into ResourceClaims for the DRA driver.
+* The DRA driver GPU capability for ``gpu.nvidia.com``, ``mig.nvidia.com``, and ``vfio.gpu.nvidia.com`` devices.
+  This capability is always enabled.
```
bmozaffa:
```suggestion
  This capability is always enabled, although VFIO passthrough is unusable without the Alpha ``PassthroughSupport`` feature gate which is off by default.
```

gpu-operator/getting-started.rst
```diff
@@ -471,6 +477,21 @@ To view all the options, run ``helm show values nvidia/gpu-operator``.
        custom driver images.
      - ``nvcr.io/nvidia``
 
+   * - ``draDriver.computeDomains.enabled``
+     - When set to ``true``, the ``GPUCluster`` stack deploys the ComputeDomain controller and kubelet plugin.
+       GPU allocation through DRA is always enabled when a ``GPUCluster`` is deployed.
```
bmozaffa: This seems a bit out of place?

gpu-operator/gpu-driver-upgrades.rst
```diff
@@ -46,6 +46,11 @@ The GPU Operator supports several methods for managing and automating this drive
    The GPU Operator only manages the lifecycle of containerized drivers.
    Drivers which are pre-installed on the host are not managed by the GPU Operator.
 
+For the technology preview ``GPUCluster`` DRA stack, the Operator treats pods with an allocated
```
bmozaffa:
```suggestion
For the ``GPUCluster`` DRA stack, the Operator treats pods with an allocated
```

gpu-operator/install-gpu-operator-air-gapped.rst
```diff
@@ -242,6 +242,29 @@ Sample of ``values.yaml`` for GPU Operator v1.9.0:
       imagePullSecrets: []
 
 
+DRA Driver Image
+================
+
+If you deploy the technology preview ``GPUCluster`` custom resource definition, mirror the DRA Driver for NVIDIA GPUs image in addition to the other GPU Operator images:
```
bmozaffa:
```suggestion
If you deploy the ``GPUCluster`` custom resource definition, mirror the DRA Driver for NVIDIA GPUs image in addition to the other GPU Operator images:
```

gpu-operator/getting-started.rst
```diff
@@ -471,6 +477,21 @@ To view all the options, run ``helm show values nvidia/gpu-operator``.
...
+       Refer to :doc:`DRA Driver for NVIDIA GPUs <dra-intro-install>` for more information.
+     - ``true``
+
+   * - ``draDriver.version``
+     - Version of the DRA Driver for NVIDIA GPUs deployed by the ``GPUCluster`` stack.
+     - ``v${dra_version}``
+
+   * - ``gpuCluster.deployCR``
+     - When set to ``true``, the chart creates the technology preview ``GPUCluster`` resource for the DRA enablement stack.
```
bmozaffa:
```suggestion
     - When set to ``true``, the chart creates the ``GPUCluster`` resource for the DRA enablement stack.
```

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -30,6 +30,10 @@ You can create one or more instances of an NVIDIA driver (``NVIDIADriver``) cust
 to specify the NVIDIA GPU driver type and driver version to configure on specific nodes.
 You can specify labels in the node selector field to control which NVIDIA driver configuration is applied to specific nodes.
 
+The technology preview ``GPUCluster`` enablement stack uses ``NVIDIADriver`` resources for an Operator-managed driver.
```
bmozaffa:
```suggestion
The ``GPUCluster`` enablement stack uses ``NVIDIADriver`` resources for an Operator-managed driver.
```

gpu-operator/install-gpu-operator-air-gapped.rst
bmozaffa:
```suggestion
If you deploy the ``GPUCluster`` custom resource definition, mirror the DRA Driver for NVIDIA GPUs image in addition to the other GPU Operator images:
```

bmozaffa (changes requested):

gpu-operator/dra-intro-install.rst
mikemckiernan: I suppose it might not matter to the kind of workload that we typically run, but it was meant as a counterpoint to the device plugin's support for preemption.

https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/#limitations

I removed it, as you suggest.

gpu-operator/dra-intro-install.rst
mikemckiernan: I'm happy to revise however you prefer...based on my reading of your note, you aren't against identifying the alpha feature gates, so I'm inclined to keep the Maturity column--but change all other rows to "Generally Available."  If that's not an outcome you want, we might need to discuss.

gpu-operator/dra-intro-install.rst
mikemckiernan: My intention is to remove this text completely.

The original goal was two-fold: provide a decoder ring between the driver terminology and the Operator terminology, and to address the nuance that could cause confusion.

gpu-operator/dra-intro-install.rst
mikemckiernan: My thinking was that this row mentions "dynamic" and that'd be too strong a contrast with "preconfigured."  But, no strong feeling and happy to swap to "preconfigured" if you think that's clearer.

gpu-operator/dra-intro-install.rst
mikemckiernan: I thought it was a good approach.  My first thought was to focus on the needs of the user and answer "why would I want to use the DRA driver?"

Regarding the "focus on still missing features..." and the result that folks might prefer to use the device plugin--isn't that an honest assessment?

What do you propose? It doesn't have to be my way.

gpu-operator/dra-intro-install.rst
mikemckiernan: Not sure, editor (?), sorry for the noise.  My pref is sentence-per-line anyway.

gpu-operator/dra-intro-install.rst
mikemckiernan: Changed, but not verbatim.  The subject of your text is the software, the subject of my proposal is the reader ("you").

gpu-operator/dra-intro-install.rst
bmozaffa: Yes, that's fine, but the way to think about it, IMO, is that "folks might prefer to use the device plugin" for "a while longer". At some point, device plugin will be deprecated and they will have to move. So giving them information to decide if they are ready to make the switch **at this point** is a good idea. However, it's a question of when to switch and not if to switch.

gpu-operator/dra-intro-install.rst
bmozaffa: Yes, thank you, it looks good now. Of course whatever is Alpha and not GA should be identified as such. It's just that the API version can sometimes be orthogonal to the feature maturity, and I wanted to make sure we avoid any confusion. It looks good now!

gpu-operator/dra-intro-install.rst
bmozaffa: I think the about feature parity is duplicated now, in two back to back sections!

gpu-operator/dra-intro-install.rst
mikemckiernan: My expectation is that this documentation is gradually updated as the DRA is gradually updated to meet parity.  The language gets stronger as development progresses.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,395 @@
...
-Overview
-*****************
+*********************************
+Comparison: DRA and Device Plugin
+*********************************
 
-With the DRA Driver for NVIDIA GPUs, your Kubernetes workloads can allocate and consume the following two types of resources:
+GPU Operator supports the DRA Driver as a more recent alternative to the NVIDIA Device Plugin for Kubernetes.
+However, some features of the DRA driver are alpha maturity and not fully supported.
+The driver does not provide feature parity with the device plugin.
```
shivamerla: @mikemckiernan this gives an impression that some key features are missing in the DRA driver. That is actually quite opposite and it has much more advanced and flexible device allocation features. Instead we should just point users to carefully review current limitations {minimum k8s version, feature gate dependencies etc}.

gpu-operator/dra-intro-install.rst
shivamerla: I also want to echo what @bmozaffa is saying and the below bullet points doesn't clearly state key differences from the API, feature and scheduling perspective. I think just the comparison table added below is sufficient for users to make a decision.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,395 @@
...
+     - Alpha ``NVMLDeviceHealthCheck`` is disabled by default.
+       gRPC health probes report kubelet plugin availability only.
+     - Reports unhealthy devices through the Kubernetes Device Plugin API.
 
-    * The corresponding nvidia-imex-* packages installed through your Linux distribution's package manager.
-    * The IMEX systemd service disabled before installing the GPU Operator (on all GPU nodes).
-      For example:
+   * - Device Injection and Managed Components
+     - Requires a Container Device Interface (CDI)-compatible runtime.
+       ``GPUCluster`` manages DRA, ComputeDomains, NVIDIA Data Center GPU Manager (DCGM), DCGM Exporter, and the DRA validator.
```
shivamerla, line 84: We have to highlight the fact that with `GPUCluster` the GPU stack is much simpler and manages DRA drivers, DCGM/Exporter and a validation pod.

gpu-operator/dra-intro-install.rst
```diff
@@ -20,256 +20,395 @@
...
+
+DRA Driver Limitations
+======================
+
+Consider the following limitations before selecting DRA driver v${dra_version}:
+
+* ``NVMLDeviceHealthCheck`` is alpha and disabled by default.
+  The gRPC health probe reports kubelet plugin availability, and DCGM provides telemetry.
+  Neither provides DRA allocation health status.
+* This release does not provide scheduler-accounted capacity sharing among independent ``ResourceClaim`` objects.
```
shivamerla: the `ConsumableShares` feature is supported as an `alpha` like other feature gates and users can explore that just like other features.

shivamerla:
@mikemckiernan I think a block diagram showing the operand stack managed by both `GPUCluster` and `ClusterPolicy` would be good at the high level showcasing the differences clearly. This can be added as a follow up as well.

PR #456, opened 2026-08-07
Release docs 26.7.0

mikemckiernan:

PR #459, opened 2026-08-10
docs: Def NVIDIA driver and migration

mikemckiernan:
Documents https://github.com/NVIDIA/cloud-native-team/issues/245

Review HTML:
- Updates for [mode](https://nvidia.github.io/cloud-native-docs/review/pr-459/gpu-operator/latest/gpu-driver-configuration.html#driver-management-modes) on the driver CRD.  The same page has the migration.
- Update to the [driver upgrade](https://nvidia.github.io/cloud-native-docs/review/pr-459/gpu-operator/latest/gpu-driver-upgrades.html#upgrades-with-the-upgrade-controller) page to move the upgrade procedure here.  I made a guess that with the enhancements, we'd want more visibility into the capabilities.  LMK if you prefer it back on the driver CRD page.

### DIFF
--- a/gpu-operator/gpu-driver-configuration.rst
+++ b/gpu-operator/gpu-driver-configuration.rst
@@ -31,14 +31,19 @@ to specify the NVIDIA GPU driver type and driver version to configure on specifi
 You can specify labels in the node selector field to control which NVIDIA driver configuration is applied to specific nodes.
 
 
-Limitations
-===========
+Driver Management Modes
+=======================
 
-* This feature is recommended for new cluster installations only.
-  Upgrades from ClusterPolicy managed drivers to NVIDIA driver custom resource managed drivers are not supported.
-  Switching from ClusterPolicy to the NVIDIA driver custom resource will cause all existing driver pods to be terminated immediately and redeployed using the new NVIDIADriver configuration.
-* You must either use the default NVIDIA driver custom resource that the Helm chart creates or create and manage your own custom NVIDIA driver custom resource.
-* You can't use ClusterPolicy and the NVIDIA driver custom resource at the same time. You can only use one or the other in a cluster.
+To use NVIDIA driver custom resources with a cluster policy custom resource, set
+``spec.driver.useNvidiaDriverCRD`` to ``true``.
+The cluster policy controller continues to manage the other GPU Operator operands, but delegates
+driver management to the NVIDIA driver custom resource controller.
+If ``spec.driver.useNvidiaDriverCRD`` is ``false``, the Operator does not reconcile
+NVIDIA driver custom resources.
+
+You can migrate a Helm installation from managing the driver with the cluster policy custom resource
+to managing it with NVIDIA driver custom resources.
+For the supported procedure, refer to `Migrating from Cluster Policy Driver Management`_.
 
 Comparison: Managing the Driver with CRD versus the Cluster Policy
 ==================================================================
@@ -87,26 +92,34 @@ About the Default NVIDIA Driver Custom Resource
 ===============================================
 
 By default, the Helm chart configures a default NVIDIA driver custom resource during installation.
-This custom resource does not include a node selector and as a result, the custom resource applies to every node in your cluster
-that has an NVIDIA GPU.
+The resource has ``spec.default: true`` and does not include a node selector.
+It acts as a fallback and applies to every GPU node that does not match the node selector of another
+NVIDIA driver custom resource.
 The Operator starts a driver daemon set and pods for each operating system version in your cluster.
 
-If you plan to configure your own driver custom resources to specify driver versions, types, and so on, then
-you might prefer to avoid installing the default custom resource.
-By preventing the installation, you can avoid node selector conflicts due to the default custom resource
-matching all nodes and your custom resources matching some of the same nodes.
+You can use the default resource and additional user-defined resources at the same time.
+A user-defined resource takes precedence on nodes that match its node selector, and the default resource
+continues to manage the remaining GPU nodes.
+User-defined resources must not select the same node.
+If they do, the affected resources report ``notReady`` with a ``ConflictingNodeSelector`` condition,
+and the Operator retains the existing driver ownership labels until you resolve the conflict.
 
-To prevent configuring the default custom resource, specify the ``--set driver.nvidiaDriverCRD.deployDefaultCR=false``
-argument when you install the Operator with Helm.
+Only one NVIDIA driver custom resource can have ``spec.default: true``.
+A default resource cannot specify ``spec.nodeSelector`` because it would not be able to act as a fallback.
+If more than one default resource exists, the Operator marks the affected resources ``notReady`` with a
+``ReconcileFailed`` condition and does not change the existing driver ownership labels on nodes.
+Delete or update the extra default resource to resume reconciliation.
 
-If the Operator is already installed with the default custom resource and you want to create your own
-driver custom resources and apply them to specific nodes, delete the default custom resource.
+To prevent the Helm chart from creating the default resource, specify the
+``--set driver.nvidiaDriverCRD.deployDefaultCR=false`` argument when you install or upgrade the Operator.
+Use this setting only if your user-defined resources select every GPU node that the Operator should manage,
+or if leaving some GPU nodes without an Operator-managed driver is intentional.
 
 .. note::
 
-   After you delete the default custom resource, your custom resources might not reconcile
-   automatically due to a known issue. Refer to the :ref:`v26.3.0 known issues <v26.3.0-known-issues>`
-   for the workaround.
+   A user-defined resource without a node selector matches all GPU nodes.
+   The resource takes precedence over the default resource and conflicts with any other resource
+   that selects one of the same nodes.
 
 
 Feature Compatibility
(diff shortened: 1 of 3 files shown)


gpu-operator/gpu-driver-configuration.rst
```diff
@@ -87,26 +92,34 @@ About the Default NVIDIA Driver Custom Resource
 ===============================================
 
 By default, the Helm chart configures a default NVIDIA driver custom resource during installation.
```
rahulait, line 94: It configures default NVIDIA driver custom resource when `driver.nvidiaDriverCRD.deployDefaultCR` is set to `true`. We should update this line to mention that.

rahulait: Ah nevermind, we do set deployDefaultCR to true by default. So that should get installed whenever nvidiadriverCRD is enabled. So we are good here.
https://github.com/NVIDIA/gpu-operator/blob/599e28368646116dc56a8018a068fbac6334cce1/deployments/gpu-operator/values.yaml#L141

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -87,26 +92,34 @@ About the Default NVIDIA Driver Custom Resource
...
+If more than one default resource exists, the Operator marks the affected resources ``notReady`` with a
+``ReconcileFailed`` condition and does not change the existing driver ownership labels on nodes.
+Delete or update the extra default resource to resume reconciliation.
 
-If the Operator is already installed with the default custom resource and you want to create your own
-driver custom resources and apply them to specific nodes, delete the default custom resource.
+To prevent the Helm chart from creating the default resource, specify the
+``--set driver.nvidiaDriverCRD.deployDefaultCR=false`` argument when you install or upgrade the Operator.
+Use this setting only if your user-defined resources select every GPU node that the Operator should manage,
+or if leaving some GPU nodes without an Operator-managed driver is intentional.
```
rahulait: Or, a user wants to create their own fallback default nvidiadriver CR (by specifying `spec.default: true` for their nvidiadriver CR).

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -87,26 +92,34 @@ About the Default NVIDIA Driver Custom Resource
...
 
 .. note::
 
-   After you delete the default custom resource, your custom resources might not reconcile
-   automatically due to a known issue. Refer to the :ref:`v26.3.0 known issues <v26.3.0-known-issues>`
-   for the workaround.
+   A user-defined resource without a node selector matches all GPU nodes.
```
rahulait:
```suggestion
   A user-defined NVIDIADriver custom resource without a node selector matches all GPU nodes.
```

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -87,26 +92,34 @@ About the Default NVIDIA Driver Custom Resource
...
+   The resource takes precedence over the default resource and conflicts with any other resource
```
rahulait:
```suggestion
   The user defined NVIDIADriver CR takes precedence over the default CR and conflicts with any other CR
```

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -140,6 +153,116 @@ Custom Driver Parameters
...
+***********************************************
+
+You can migrate an existing Helm installation to NVIDIA driver custom resource management through
+the controlled driver upgrade flow.
+During the migration, the Operator assigns each GPU node to an NVIDIA driver custom resource and uses the
+driver upgrade controller to replace the previous cluster policy managed driver pod on each node.
+
+When you migrate from a GPU Operator release earlier than v26.7.0, perform two Helm upgrades.
+First, upgrade to v26.7.0 while retaining cluster policy driver management.
+Then, upgrade the same release again to enable NVIDIA driver custom resource management.
```
rahulait:
```suggestion
Then, upgrade to the same release again enabling NVIDIADriver to enable NVIDIA driver custom resource management.
```

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -140,6 +153,116 @@ Custom Driver Parameters
...
+This sequence starts the controller that supports controlled migration before changing driver ownership.
+Do not upgrade from an earlier release and enable NVIDIA driver custom resource management in the same
```
rahulait, line 180:
```suggestion
Do not upgrade from an earlier release (using clusterpolicy for driver management) and enable NVIDIA driver custom resource management in the same
```

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -140,6 +153,116 @@ Custom Driver Parameters
...
+***********************************************
+Migrating from Cluster Policy Driver Management
+***********************************************
+
+You can migrate an existing Helm installation to NVIDIA driver custom resource management through
+the controlled driver upgrade flow.
+During the migration, the Operator assigns each GPU node to an NVIDIA driver custom resource and uses the
+driver upgrade controller to replace the previous cluster policy managed driver pod on each node.
+
+When you migrate from a GPU Operator release earlier than v26.7.0, perform two Helm upgrades.
```
rahulait:
```suggestion
When you migrate from a GPU Operator release earlier than v26.7.0, perform two Helm upgrades. Clusters on or after v26.7.0 don't need the two step upgrade and can switch from clusterpolicy to nvidiadriver in a single upgrade.
```

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -140,6 +153,116 @@ Custom Driver Parameters
...
+* Your workloads can tolerate the disruption configured by the driver upgrade policy.
+  For more information, refer to :ref:`gpu-driver-upgrades`.
+
+#. Identify the Helm release name:
+
+   .. code-block:: console
+
+      $ helm list -n gpu-operator
+
+#. If your current GPU Operator release is earlier than v26.7.0, upgrade to the target release while
```
rahulait:
```suggestion
#. If your current GPU Operator release is earlier than v26.7.0 (and not using nvidiadriver CRs), upgrade to the target release while
```

repo.toml
rahulait, line 229: Does this need an update as well?

gpu-operator/gpu-driver-configuration.rst
mikemckiernan: There's some precedent for dropping "custom" when the context is high that the subject is a custom resource.

Here's an example from upstream: https://kubernetes.io/docs/concepts/extend-kubernetes/operator/#example

Step 1 uses "A custom resource named SampleDB..." and step 4: "...find out what SampleDB resources are configured."

I anticipate your goal is precision.  My concern is that this level of repetition is harmful to human readers and likely inconsequent to agentic readers.

gpu-operator/gpu-driver-configuration.rst
mikemckiernan: I don't dispute that what you propose is technically accurate, but I find "...again enabling NVIDIADriver to enable NVIDIA driver..." difficult to read.

My proposal:

```
When you migrate from a GPU Operator release earlier than v26.7.0, perform two Helm upgrades:

* First, upgrade to v26.7.0 while retaining cluster policy driver management.
* Afterward, upgrade to the same release again.
  This time, enable NVIDIA driver custom resource management.

  The exact commands and arguments are shown in the following procedure.
```

gpu-operator/gpu-driver-configuration.rst
mikemckiernan: My instinct is to nuke the whole "Do not..." para.  I think the instruction that precedes it and the procedure that follows are abundantly clear.

gpu-operator/gpu-driver-configuration.rst
rahulait: Sure, sounds good to me.

gpu-operator/gpu-driver-configuration.rst
mikemckiernan: Two things:

- First, I'd prefer to avoid introducing the two-upgrade process for < 26.7.0 and then immediately provide the counterpoint circumstance before a human reader even reads the rationale for the two-upgrade process.
- Second, the conditional in the procedure seems sufficiently clear: `#. If your current GPU Operator release is earlier than v26.7.0, upgrade...`.

My concern is that if we try to handle "< 26.7.0 and >= 26.7.0" sentence-by-sentence, that we're taxing the attention span of our human readers.

gpu-operator/gpu-driver-configuration.rst
rahulait: Ack, sounds good then.

gpu-operator/gpu-driver-configuration.rst
mikemckiernan: This whole procedure is under a "Migrating from Cluster Policy Driver Management" title, but I'll yield because we still have some human readers who sometimes do not take care.

gpu-operator/gpu-driver-configuration.rst
rahulait: Sure, sounds good to me.

repo.toml
mikemckiernan: It does and thank you for asking.  I plan to cover that in the broader and shallower PR https://github.com/NVIDIA/cloud-native-docs/pull/456/changes#diff-f8fd821ad281f218ad3d80217856da5048e19a822509822c6d6ce9c2032e9660

gpu-operator/gpu-driver-configuration.rst
```diff
@@ -140,6 +156,122 @@ Custom Driver Parameters
...
+   .. code-block:: console
+
+      $ helm upgrade <release-name> nvidia/gpu-operator \
+          -n gpu-operator \
+          --version=${version} \
+          --reuse-values \
+          --disable-openapi-validation \
+          --set operator.upgradeCRD=true \
+          --set driver.nvidiaDriverCRD.enabled=false \
+          --set driver.nvidiaDriverCRD.deployDefaultCR=false \
```
rahulait, line 218: Do we need to set these (as they are the same as default value in chart)? We can avoid specifying them to avoid confusion where one might think these are always required.

rahulait (approved):

PR #462, opened 2026-08-11
docs: Single install procedure

mikemckiernan:
Review HTML:
- A single (somewhat more streamlined) install procedure: https://nvidia.github.io/cloud-native-docs/review/pr-462/confidential-containers/latest/confidential-containers-deploy.html
- Not actually HTML, but the [previous install-quickstart](https://github.com/mikemckiernan/cloud-native-docs/blob/mmck-fix-coco-one-procedure/repo.toml#L220) is now a redirect to the single install procedure.


confidential-containers/confidential-containers-deploy.rst
```diff
@@ -18,171 +18,99 @@
...
-
-      NAME      STATUS   ROLES    AGE   VERSION
-      node-01   Ready    <none>   10d   v1.34.0
-
-**Success criteria:** All nodes you intend to use for Confidential Container workloads appear in the output above.
-
-After all your desired nodes are labeled, you can continue to the next step to install Kata Containers.
+         .. code-block:: console
 
+            $ kubectl label node <node-01> <node-02> nvidia.com/gpu.workload.config=vm-passthrough
```
fidencio, line 74: We could add to the .values that we provide something like:
```
nodeSelector:
  nvidia.com/gpu.workload.config: "vm-passthrough"
```

This would ensure that kata only gets installed in the nodes that are labelled. However, before doing so we'd need to ensure it QA does a pass on this.

PR #465, opened 2026-08-12
chore: Address rhcos and rhel image names

mikemckiernan:


rahulait (approved):
LGTM. I'll wait for another approval before merging it in.

PR #466, opened 2026-08-12
docs: Use kata values file in quickstart install

fidencio:
Align the quickstart with the detailed guide by showing kata-nvidia-gpu-values.yaml and installing with -f instead of inline --set.


zvonkok (approved):

fidencio:
While this adds duplication, this fixes the crucial missing information from the Quickstart that makes the installation fail for anyone who ends up following the guide.

Later PRs can fix, simplify, unify, beautify, and clarify. For now, let's focus on getting the fix in.

PR #473, opened 2026-08-20
docs: KubeVirt with DRA

mikemckiernan:
Review HTML: https://nvidia.github.io/cloud-native-docs/review/pr-473/gpu-operator/latest/gpu-operator-kubevirt-dra.html

### DIFF
--- a/gpu-operator/gpu-operator-kubevirt-dra.rst
+++ b/gpu-operator/gpu-operator-kubevirt-dra.rst
@@ -1,0 +1,577 @@
+.. license-header
+  SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: Apache-2.0
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+.. headings (h1/h2/h3/h4) are # * = -
+
+.. _gpu-operator-kubevirt-dra:
+
+##################################
+GPU Operator with KubeVirt and DRA
+##################################
+
+KubeVirt can attach NVIDIA GPUs to virtual machines by using Kubernetes Dynamic Resource Allocation (DRA).
+The NVIDIA DRA driver advertises GPUs and their attributes in ``ResourceSlice`` objects.
+A ``ResourceClaimTemplate`` selects devices, Kubernetes schedules the virtual machine on a node that can access those devices, and the DRA driver prepares the allocation for VFIO passthrough.
+
+This workflow uses a GPU Operator-managed ``GPUCluster`` resource and DRA operands.
+For passthrough with the sandbox device plugin, or for NVIDIA vGPU mediated devices, refer to :ref:`gpu-operator-kubevirt`.
+
+****************************
+Understanding GPU Assignment
+****************************
+
+With DRA, the device request specifies the required device type and attributes.
+Kubernetes allocates devices that satisfy the request and schedules the virtual machine on the node that advertises them.
+The DRA driver then binds each allocated GPU to ``vfio-pci`` while the claim is prepared and restores it when the claim is unprepared.
+On supported systems, Fabric Manager partitioning can additionally constrain a multi-GPU claim to a valid NVSwitch fabric partition.
+
+GPU Operator discovers GPU nodes and manages operand placement.
+
+******************************************
+Assumptions, Constraints, and Dependencies
+******************************************
+
+* The examples allocate full GPUs from the ``vfio.gpu.nvidia.com`` DeviceClass.
+  MIG device allocation is outside the scope of this procedure.
+* An allocated GPU cannot be used by another workload until the claim is unprepared.
+* Host services and workloads must release open handles to a GPU before it can be bound to ``vfio-pci``.
+* KubeVirt supports exactly one allocated device for each named request.
+  Use one named request for each GPU, as shown in this procedure.
+* Do not mix DRA and device-plugin GPU entries in the same virtual machine.
+* Virtual machines with passthrough GPUs cannot be live migrated.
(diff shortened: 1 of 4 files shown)


gpu-operator/gpu-operator-kubevirt-dra.rst
```diff
@@ -0,0 +1,577 @@
...
+   .. code-block:: console
+
+      $ sudo systemctl stop nvidia-fabricmanager
+
+#. In the Fabric Manager configuration file, configure the following settings:
+
+   .. code-block:: ini
+
+      FABRIC_MODE=1
+      FABRIC_MODE_RESTART=1
```
varunrsekar: nit: remove this
```suggestion
```

gpu-operator/gpu-operator-kubevirt-dra.rst
```diff
@@ -0,0 +1,577 @@
...
+
+#. On the host, list the Fabric Manager partitions and their status:
+
+   .. code-block:: console
+
+      $ /run/nvidia/fmpm \
+          --unix-domain-socket /run/nvidia-fabricmanager/socket \
+          -l
+
+   Confirm that the partition selected for the claim reports ``isActive: 1``.
```
varunrsekar: This does not exist. 
You may use this example but : https://github.com/NVIDIA/go-nvfm/blob/main/examples/list-partitions/main.go
The steps are:
```
git clone https://github.com/NVIDIA/go-nvfm && cd go-nvfm
make example-list-partitions
kubectl cp ./list-partitions -n gpu-operator <driver-pod>:/run/nvidia/list-partitions
kubectl exec -n gpu-operator <driver-pod> -- bash -c "NVFM_UNIX_SOCKET_PATH=/run/nvidia-fabricmanager/socket /run/nvidia/list-partitions"
```

But I don't think we should include this validation step. cc: @shivamerla

gpu-operator/gpu-operator-kubevirt-dra.rst
```diff
@@ -0,0 +1,577 @@
...
+
+      $ kubectl patch gpucluster gpu-cluster \
+          --type=merge -p '{
+              "spec":{
+                "draDriver":{
+                  "computeDomains":{"enabled":false},
+                  "featureGates":{
+                    "PassthroughSupport":true,
+                    "DeviceMetadata":true,
+                    "FabricManagerPartitioning":true  # For NVLink 5 or later systems with NVSwitch-managed fabric.
```
bmozaffa, line 201: This looks like a command to copy and paste... but I think the comment would break it?

gpu-operator/gpu-operator-kubevirt-dra.rst
```diff
@@ -0,0 +1,577 @@
...
+Assumptions, Constraints, and Dependencies
+******************************************
+
+* The examples allocate full GPUs from the ``vfio.gpu.nvidia.com`` DeviceClass.
+  MIG device allocation is outside the scope of this procedure.
+* An allocated GPU cannot be used by another workload until the claim is unprepared.
+* Host services and workloads must release open handles to a GPU before it can be bound to ``vfio-pci``.
+* KubeVirt supports exactly one allocated device for each named request.
+  Use one named request for each GPU, as shown in this procedure.
+* Do not mix DRA and device-plugin GPU entries in the same virtual machine.
```
bmozaffa, line 56: I think we need another warning here, perhaps with input from engineering, about our race condition between GPU and VFIO allocation

bmozaffa (changes requested):

gpu-operator/gpu-operator-kubevirt-dra.rst
shivamerla: @varunrsekar this could be useful for savvy admins who want to check that the partitions are activated. Anyway the repo is public, they will report if any issues with it. We cannot use this directly? https://github.com/NVIDIA/Fabric-Manager-Client

gpu-operator/gpu-operator-kubevirt-dra.rst
mikemckiernan: I believe it does, but I was willing to include it because:

A) I didn't want two copy-paste commands that are identical except for one line--confusing, imo.
B) If I add "For NVLink5 or later...add the "FabricManagerPartitioning":true feature gate" above or below the code block, I fear it'll get missed.  Conversely, if I put a "If you host isn't, remove..." above or below, also might not get read and I'm unsure if that was tested or could cause a problem.

My thinking is that _nearly_ cut-n-paste that is clear and has the added bonus of failing-command-attention-getting behavior was the right choice.

If you have a suggestion, please share.

gpu-operator/gpu-operator-kubevirt-dra.rst
mikemckiernan: Very grateful to you both for the review and letting me know the client isn't automatically installed. (Why?)

I revised to group the two verification/informational commands under an optional step that mentions "requires additional software."  It's honest, hopefully prevents surprise, and still helps a determined admin gain more knowledge.  Well, imo.

If it's still problematic, plmk and thanks again!

gpu-operator/gpu-operator-kubevirt-dra.rst
mikemckiernan: @shivamerla, @varunrsekar, plmk if you agree with Babak's concern.  I believe Babak is referring to the second bullet from the [KubeVirt VFIO limitations and considerations](https://dra-driver-nvidia-gpu.sigs.k8s.io/docs/guides/gpu-allocation/kubevirt-vfio-gpu-passthrough/#limitations-and-considerations) in the DRA driver docs.

shivamerla: Yes Mike, better to highlight that limitation here as well as mixed workloads are allowed now by default with DRA.

gpu-operator/gpu-operator-kubevirt-dra.rst
varunrsekar: > We cannot use this directly? https://github.com/NVIDIA/Fabric-Manager-Client

No there's a couple of issues with it:
1. A bug that prevents it from working with unix socket
1. Building the binary require installing `jsoncpp` and the FM SDK 

The go-nvfm examples have no dependeicies so would be easier to build.

>  this could be useful for savvy admins who want to check that the partitions are activated

And yeah its good if they know what they're doing. We're injecting a binary into the driver container and running something inside it so I was skeptical. Maybe we should come up with instructions using a debug container?

varunrsekar: @mikemckiernan FYI The verification steps require the client to be run inside the driver container to work.

PR #474, opened 2026-08-20
fix: k get po command

mikemckiernan:


fidencio (approved):
lgtm, thanks @mikemckiernan!

PR #479, opened 2026-08-25
coco docs: clarify platforms page and agent policy

mikemckiernan:
Rename the supported platforms page to include software components, and document attaching a Kata agent security policy for attested production workloads.

PR #480, opened 2026-08-25
fix: clarify prereq, installed, post-inst

mikemckiernan:


confidential-containers/supported-platforms.rst
```diff
@@ -75,16 +75,16 @@ NVIDIA GPUs
...
 
-   * - Category
+   * - CPU Platform
      - TEE
-     - Operating System
-     - Kernel Version
+     - Host Operating System
+     - Host Kernel Version
    * - AMD Genoa / Milan
      - AMD SEV-SNP
```
mikemckiernan, line 89: @manuelh-dev , I think you proposed removing this information or moving it.  I took a look at the page and my instinct is to keep it:

- It's not a disaster as it is. (Yay!)
- I suspect the CPUs and technology are not yet widely prevalent and highlighting the required CPU features seems important.  (For now.)
- I suspect it meets expectations--that there would be a matrix for CPUs.

mikemckiernan (commented):
More of a question.

confidential-containers/supported-platforms.rst
manuelh-dev: @mikemckiernan - my suggestion was to move the Host Kernel Version and Host Operating System fields, specifically, down into the list below - but after your re-organization below, I think we are good. Agree that we can keep this information here

confidential-containers/supported-platforms.rst
```diff
@@ -128,19 +161,27 @@ Supported Software Components
...
+The following components and interfaces are not installed by the :doc:`Quickstart Install <install-quickstart>` or :doc:`Detailed Install Guide <confidential-containers-deploy>`.
+
+.. flat-table::
+   :header-rows: 1
+
+   * - Interface or Component
+     - Version
+     - When It Is Needed
+     - How It Is Provided
    * - `Key Broker Service (KBS) protocol <https://confidentialcontainers.org/docs/attestation/>`__
```
manuelh-dev, line 177: @fitzthum @fidencio - I think we need a bit of a refinement here. I think the KBS protocol is installed implicitly by shipping the attestation agent, CDH in our guests. So, I think here we want to say that trustee is not installed - and provide a specific trustee version to install, instead.

Then, in above listing for Kata Containers we can say that this comes with a guest using the 0.4.0 KBS protocol. This may make it easier here in turn because we can say to install a trustee version using that same protocol version (or, a protocol version that is compatible, if a patch version increment of the protocol version means that both are still interoperable).

manuelh-dev:
> More of a question.

I really like this, thank you for these changes! I left one comment for discussion.

confidential-containers/supported-platforms.rst
```diff
@@ -128,19 +161,27 @@ Supported Software Components
...
-     - ${kata_version} (installed with ``kata-deploy`` Helm chart)
+Separately Deployed Components and Interfaces
+=============================================
+
+The following components and interfaces are not installed by the :doc:`Quickstart Install <install-quickstart>` or :doc:`Detailed Install Guide <confidential-containers-deploy>`.
+
+.. flat-table::
+   :header-rows: 1
+
+   * - Interface or Component
```
manuelh-dev, line 173: @mikemckiernan - if or as we make this change, we could also list genpolicy - we should then merge https://github.com/NVIDIA/cloud-native-docs/pull/479 after this one.

manuelh-dev: Let's either merge this one first, then adjust #479, or vice versa.

confidential-containers/supported-platforms.rst
dcmiddle, line 62: Why do we call out HGX here but not the other rows? 
My understanding is that HGX is a platform reference design (inclusive of GPU, switch, CPU) productized by various OEMs. Seems like it should be on all of these rows or none of them. I may certainly be ignorant of other context around HGX though.

confidential-containers/supported-platforms.rst
```diff
@@ -115,11 +115,44 @@ For additional resources on machine setup:
...
+The supported ``kata-deploy`` Helm chart installs the guest OS, guest kernel, OVMF, and QEMU artifacts listed in the following table as part of Kata Containers.
+You do not supply or install these artifacts individually.
+
+.. flat-table::
+   :header-rows: 1
+
+   * - Artifact
+     - Release/Version
    * - Guest OS
      - Distroless
```
dcmiddle, line 157: @manuelh-dev 
Is there something more specific we can list here? 
I think it's a specific nvidia flavored chiseled os including nvrc.
Is there an upstream kata link to something like a manifest?

manuelh-dev: @fidencio @zvonkok - I don't think there is. But for our distoless guest OS we could list the NVRC version and NVIDIA drivers included?

confidential-containers/supported-platforms.rst
manuelh-dev: I haven't been touching this. I was under the assumption that this was previously sorted out. Let's circle back internally here.

confidential-containers/supported-platforms.rst
fitzthum: Yeah, for various reasons it is best to use Trustee and guest-components from the same release, at least this is the safest advice to give people. More concretely, when you go to deploy Trustee, make sure you are using the version that goes with the CoCo release. 

It's good to list the Trustee version here since it might not be obvious to people how to find the Trustee version from a particular CoCo/Kata release (not to mention a release of our RA). For reference, the Trustee [releases](https://github.com/confidential-containers/trustee/releases) say which version of CoCo they go with. Currently it trails by one.

There are some other errata involving versions that I will list here for the sake of clarity. First, the Trustee Helm chart and docker compose do not do a great job of pinning the versions of the various components. This hopefully be fixed soon as we update the Trustee release process. As is, you basically always get the latest version of the Trustee components, no matter which version fo the repo you check out, unless you manually specify which images to use. Technically, we should tell people to do that atm.

Trustee more generally is working to better version all of its interfaces. At some point we hope to cut a major release, which will be the basis for backwards compatibility. Before we can do that, we need to properly version all of the interfaces of Trustee. The KBS protocol is one of those interfaces, but this doesn't capture all the things that could change (like the format of the hw evidence itself, the API of the policy, the config files, etc). So, as mentioned, the KBS protocol version isn't something that we really need to elevate. Instead, it's better to thinking in terms of Trustee versions and to have an exact match for now, although in practice that last several releases can work together without issues.

confidential-containers/supported-platforms.rst
```diff
@@ -28,7 +28,7 @@ This page is relevant to the following users:
 
 * The :ref:`Hardware IT Administrator <coco-persona-hardware-it-administrator>` uses the hardware tables to confirm that the selected CPU and GPU are validated for Confidential Computing before configuring the system.
```
Hema-Bontha-NV, line 29: one potential option is to add a pointer to the NVIDIA Certification w/ CC that essentially solves for HW CC pre-requisites and performance pre-requisites for CC (and the stack is a K8S stack). Will find that documentation and add link here, if we think it makes sense.

confidential-containers/supported-platforms.rst
Hema-Bontha-NV: it should be called out as HGX for all the rows or have a generic disclaimer above common to all rows. What we need convey is that the qualification is with x86 CPU + GPU and not a DGX variant (which as SW dependencies on very specific control plane components, that we do not test with yet).

confidential-containers/supported-platforms.rst
```diff
@@ -75,16 +75,16 @@ NVIDIA GPUs
     For both single and multi GPU Passthrough, all GPUs on the host must be configured for Confidential Computing and all GPUs must be assigned to one Confidential Container virtual machine.
```
Hema-Bontha-NV, line 75: One thing that is not clear in the documentation is weather or not we support multi-GPU with different VM Shapes for Multi-GPU support other than 8 GPUs (meaning VM shapes with 2, or 4  GPUs). I understand there is a dependency on who should be able to make these configurations (tenant vs. Operator) and how to leverage Fabric manager to get this working. So we need to evaluate where and how we state that in the documentation.

confidential-containers/supported-platforms.rst
Hema-Bontha-NV: One aspect on the Gen Policy tool related documentation is that it is being treated as an optional component today (atleast when we look at what Red Hat does with their solution). So should we go to the depth of explaining what is the function of the Gen-policy tool and init.data and what options exist to get to that state.

confidential-containers/supported-platforms.rst
```diff
@@ -128,19 +161,27 @@ Supported Software Components
...
+Separately Deployed Components and Interfaces
+=============================================
+
+The following components and interfaces are not installed by the :doc:`Quickstart Install <install-quickstart>` or :doc:`Detailed Install Guide <confidential-containers-deploy>`.
+
+.. flat-table::
+   :header-rows: 1
+
+   * - Interface or Component
+     - Version
```
Hema-Bontha-NV, line 174: Are we going to support a very specific version of each component or will we support N-1 release as well ? And if there is a support for N-1 release for any component, then we need to capture two things:
1. an upgrade path for components (if we support it) or explicitly state that we do not support upgrade paths and provide guidance on how new versions should be installed and what the expectations should be. 
2. is there a notion of compatible versions of different components. What i mean is if Trustee needs to be a specific version, then does it impact which version of KBS or Kata deploy or other components should be used. Need to call out the dependencies (if any).

confidential-containers/supported-platforms.rst
manuelh-dev: Thanks, deferring this to a future PR in which we better describe these version dependencies.

confidential-containers/supported-platforms.rst
manuelh-dev: I thought that was the goal of PR #479, to introduce the genpolicy tool - since there was no comments on this PR, I am carefully resolving this comment. We can improve descriptions around genpolicy as required in a follow-up PR.

confidential-containers/supported-platforms.rst
manuelh-dev: Agree'd, but a discussion unrelated to this PR.

confidential-containers/supported-platforms.rst
manuelh-dev: Good points. Deferring this to a separate PR. This goes beyond this PR

confidential-containers/supported-platforms.rst
manuelh-dev: It does make sense. Since this is a comment out of scope for this PR, we can do this in a separate PR

confidential-containers/supported-platforms.rst
manuelh-dev: Agree'd - deferring to another PR - this feedback point seems out of scope for the goals of this PR.

manuelh-dev (approved):

PR #491, opened 2026-09-09
OpenShift Virtualization: add hyperconverged v1 API

nunnatsa:
OpenShift virtualization v4.22.0 introduced the new v1 API version for the `HyperConverged` kind. This API version is not backward compatible with the older `v1beta1` API version.

The `v1beta1` API version will be deprecated, and later removed from OpenShift Virtualization.

This PR adding the v1 API tab, whenever the v1beta1 format is already used in the documents.

Note: In OpenShift Virtualization v4.22.x, because of the unchangeable way Kubernetes prioritizes API versions, using a non-versioned name of the HyperConverged custom resource, i.e. `hyperconverged` or `hco`, will make Kubernetes to use the v1 API version. It is recomanded to use the fully versioned named for v1beta1, i.e.
`hyperconvergeds.v1beta1.hco.kubevirt.io`.

### DIFF
--- a/openshift/openshift-virtualization.rst
+++ b/openshift/openshift-virtualization.rst
@@ -103,18 +103,48 @@ Prerequisites
 
 * `Install the OpenShift Virtualization Operator <https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html/virtualization/installing>`__.
 * `Install the virtctl client <https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html/virtualization/getting-started#installing-virtctl_virt-using-the-cli-tools>`__.
-* Starting with OpenShift Virtualization 4.12.3 and 4.13.0, set the ``disableMDevConfiguration`` feature gate:
+* You must disable OpenShift Virtualization mediated device management to prevent conflicts with the GPU Operator:
 
-  .. code-block:: console
+  .. tab-set::
 
-     $ kubectl patch hyperconverged -n openshift-cnv  kubevirt-hyperconverged --type='json' -p='[{"op": "add", "path": "/spec/featureGates/disableMDevConfiguration", "value": true}]'
+    .. tab-item:: OpenShift Virtualization v4.23.0 or newer
+      :selected:
 
-  *Example Output*
+      Set the ``spec.virtualization.mediatedDevicesConfiguration.enabled`` field to ``false``:
 
-  .. code-block:: output
+      .. code-block:: console
 
-     hyperconverged.hco.kubevirt.io/kubevirt-hyperconverged patched
+        $ oc patch hyperconverged -n openshift-cnv kubevirt-hyperconverged --type='merge' \
+          -p='{"spec":{"virtualization":{"mediatedDevicesConfiguration":{"enabled": false}}}}'
 
+      *Example Output*
+
+      .. code-block:: output
+
+        hyperconverged.hco.kubevirt.io/kubevirt-hyperconverged patched
+
+    .. tab-item:: Earlier versions of OpenShift Virtualization
+
+      Set the ``disableMDevConfiguration`` feature gate:
+
+      .. code-block:: console
+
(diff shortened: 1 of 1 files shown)


openshift/openshift-virtualization.rst
```diff
@@ -103,18 +103,48 @@ Prerequisites
 
 * `Install the OpenShift Virtualization Operator <https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html/virtualization/installing>`__.
 * `Install the virtctl client <https://docs.redhat.com/en/documentation/openshift_container_platform/latest/html/virtualization/getting-started#installing-virtctl_virt-using-the-cli-tools>`__.
-* Starting with OpenShift Virtualization 4.12.3 and 4.13.0, set the ``disableMDevConfiguration`` feature gate:
+* Starting with OpenShift Virtualization 4.12.3 and 4.13.0, disable MDev Configurations
```
mikemckiernan:
```suggestion
* You must disable mediated device virtualization to prevent conflicts with the Operator:
```

- These releases seem to be in the distant past.
- That's an AI proposed reason, please check the true rationale.
- Rereading this now, this seems like it might more accurately be a step in a procedure than a prereq. (Up to you.)

openshift/openshift-virtualization.rst
```diff
@@ -103,18 +103,48 @@ Prerequisites
...
+      :selected:
 
-  *Example Output*
+      Set the ``spec.virtualization.mediatedDevicesConfiguration.enabled`` field to ``false``:
 
-  .. code-block:: output
+      .. code-block:: console
 
-     hyperconverged.hco.kubevirt.io/kubevirt-hyperconverged patched
+        $ kubectl patch hyperconverged -n openshift-cnv kubevirt-hyperconverged --type='merge' \
```
mikemckiernan: Not `oc`?  Up to you, no matter to me.

openshift/openshift-virtualization.rst
```diff
@@ -103,18 +103,48 @@ Prerequisites
...
+
+        hyperconverged.hco.kubevirt.io/kubevirt-hyperconverged patched
+
+      .. note::
+
+         in OpenShit Virtualization v4.22.x, use the full versioned name for the hyperconverged custom resource:
+
+         .. code-block:: console
+
+            $ kubectl patch hyperconvergeds.v1beta1.hco.kubevirt.io -n openshift-cnv  kubevirt-hyperconverged --type='json' -p='[{"op": "add", "path": "/spec/featureGates/disableMDevConfiguration", "value": true}]'
```
mikemckiernan: I feel for OCP support folks that they need to support (test?) all these combinations.  Is there truly no more limited support matrix?

Because all three commands are unique, I think it'll present clearest if this is a labelled tab too.  Something like "OpenShift Virtualization v4.22.x"

mikemckiernan (commented):
Thank you for the update!  PLMK what I can clarify.

openshift/openshift-virtualization.rst
nunnatsa: yes...

But we're in the middle of a process to replace the API. this is the price I guess.

openshift/openshift-virtualization.rst
lyarwood: It's still required, otherwise OpenShift Virtualization itself will try to manage these devices. I'd reword the suggested change to highlight this.

```
You must disable OpenShift Virtualization mediated device management to prevent conflicts with the GPU Operator:
```

nunnatsa:
Thanks for the review @mikemckiernan. Fixed.

openshift/openshift-virtualization.rst
nunnatsa: Thanks @lyarwood !

nunnatsa:
@mikemckiernan, could you please take a look?

mikemckiernan (approved):
Thanks for the update and the revision!

PR #494, opened 2026-09-14
coco docs: improvements for version pinning and minor fixes

manuelh-dev:
A few changes:
- pin Trustee and guest-components to v0.21.0 to match this reference architecture (clone that tag instead of `main`; drop the KBS protocol 0.4.0 row).
- improve the supported-platforms listing: NVRC, guest GPU driver, guest-components in the distroless guest, and guest kernel / QEMU from Kata 4.1.0.
- pin genpolicy to 4.1.0 (compatible, includes a CVE fix).
- restore `--wait` on the detailed kata-deploy install.
- correct kata-deploy NFD settings: the sample values file and the GPU Operator Kata `--set` used `nfd.enabled`, which only applies to the GPU Operator chart. Use `node-feature-discovery.enabled` so kata-deploy *actually* turns NFD off and does not lean on a default.
- record changes as post-release documentation updates in the CoCo 1.1.0 and GPU Operator 26.7.0 release notes.


confidential-containers/supported-platforms.rst
manuelh-dev, line 142: 26.3 and higher vs 26.7 aligned with Kata + GPU operator doc - to discuss whether we want a change here.

confidential-containers/attestation.rst
```diff
@@ -68,10 +68,11 @@ Refer to the upstream `Confidential Containers documentation <https://confidenti
 Quickstart
 ==========
 
-This page walks you through standing up a development Trustee instance with Docker Compose, installing the Key Broker Service (KBS) client tool, and sending a sample resource request to confirm the system is reachable.
+This page walks you through standing up a development Trustee ${trustee_version} instance with Docker Compose, installing the Key Broker Service (KBS) client tool, and sending a sample resource request to confirm the system is reachable.
 It runs on a standalone Linux host and does not require a Kubernetes cluster or the Confidential Containers runtime.
 
 The goal is to give you a working local attestation backend and a client to interact with it before you wire Trustee into a Confidential Containers workload.
+This architecture validates Trustee ${trustee_version} with guest-components ${guest_components_version} in the Kata ${kata_version} guest.
```
mikemckiernan: I'm ambivalent about having the version information inline here and on the support matrix page.  It's not wrong to have the information here, I'm just pondering if it dilutes the value of the support page or is limiting if one page adds "and higher", but the other occurrence does not.

Again, not saying it's wrong and not asking you to remove it, just sharing that I noticed version info in two places with this update.

confidential-containers/attestation.rst
```diff
@@ -142,6 +143,10 @@ Start the Trustee containers in the background.
    On first run, ``docker compose up -d`` pulls the KBS, AS, and RVPS images before starting them.
    This step can take several minutes. The command returns after the containers start. The services may need an additional few seconds to become ready to accept requests.
 
+   The compose file at Trustee ${trustee_version} still names ``:latest`` images.
+   After you clone that tag, the configuration files match this architecture.
+   The container images follow whatever ``:latest`` resolves to unless you pin those ``image:`` lines to a digest or commit tag.
```
mikemckiernan: I find this a little warty and distracting.  It's consuming my attention because I'm wondering whether this is ever truly an issue. If end users encounter issues with using `latest`, then I feel like the best fix is to modify the compose file rather than ask every end user to modify the file--assuming I understood correctly.

If no one is ever bit by latest wandering to a newer release, or n==1, I'd say omit this info.   I asked codex for some research on this topic and it seems like digest is the precise answer--replacing with `0.21.0` supposedly wouldn't work for KBS, AS and RVPS.  If we need to keep this info, rather than the research, would someone prefer to `docker compose up -d --build`?

confidential-containers/confidential-containers-deploy.rst
```diff
@@ -201,7 +202,7 @@ The minimum required version is ${kata_version}.
    .. note::
 
       Both ``kata-deploy`` and the GPU Operator deploy Node Feature Discovery (NFD) by default.
-      The install command includes ``--set nfd.enabled=false`` to prevent ``kata-deploy`` from deploying NFD.
+      The values file sets ``nfd.enabled: false`` to prevent ``kata-deploy`` from deploying NFD.
```
mikemckiernan: ugh, thank you!

confidential-containers/supported-platforms.rst
mikemckiernan: I suspect that "and higher" is practical, but now out of place in comparison with all the other pinning.  If we're committed to the pinning, my $0.02 is to remove "and higher."

repo.toml
```diff
@@ -215,7 +215,7 @@ docs_root = "${root}/confidential-containers"
 project = "confidential-containers"
 name = "NVIDIA Confidential Containers Architecture"
 version = "1.1.0"
-source_substitutions = { kata_version = "4.0.0", gpu_operator_version = "v26.3.1", gpu_operator_minor_version = "26.3" }
+source_substitutions = { kata_version = "4.0.0", gpu_operator_version = "v26.3.1", gpu_operator_minor_version = "26.3", genpolicy_version = "4.1.0", nvrc_version = "v0.1.5", guest_nvidia_driver_version = "595.58.03", guest_kernel_version = "6.18.35", qemu_version = "11.0.1", trustee_version = "v0.21.0", guest_components_version = "v0.21.0", kbs_client_tag = "sample_only-258ea4acb7b9bd865fce5c63a539f2120dba8298-x86_64" }
```
mikemckiernan: If Glenn can say he used 26.7, I'd align with the GPU Op doc.

mikemckiernan (approved):
One small request, can you update the release-notes.rst with something like a [Post-Release Documentation Updates](https://github.com/NVIDIA/cloud-native-docs/blob/main/gpu-operator/release-notes.rst?plain=1#L191) section?  My concern is to prevent gaslighting a reader who notices changes w/o a record other than commit history.

confidential-containers/confidential-containers-deploy.rst
fidencio: I think we should rely on Kata Containers to set the NFD to true, instead of doing so with the GPU Operator.

manuelh-dev: Would it be better to make this ownership change with the next proper version update (e.g. minor version update)? Addressing this would change deployment behavior and lifecycle ownership; for now, this PR was intended to only make correctness fixes without changing the validated installation flow.

confidential-containers/attestation.rst
dcmiddle: I like the DRY principle. Just mention versions in one place. 
Not mentioning them here also has the tiny advantage of reducing word count which helps readability.

confidential-containers/attestation.rst
```diff
@@ -114,12 +115,12 @@ Step 1: Install Trustee with Docker Compose
 -------------------------------------------
 
 Installing Trustee with Docker Compose is the recommended install path.
```
dcmiddle, line 116: We recommend the helm chart now.
https://github.com/confidential-containers/trustee/#helm-chart

confidential-containers/attestation.rst
```diff
@@ -114,12 +115,12 @@ Step 1: Install Trustee with Docker Compose
...
-Clone the upstream Trustee repository.
+Clone the upstream Trustee repository at ${trustee_version}.
 The repository ships with a ``docker-compose.yml`` that wires KBS, the Attestation Service, and the Reference Value Provider Service together.
 
 .. code-block:: console
 
-   $ git clone https://github.com/confidential-containers/trustee.git && cd trustee
+   $ git clone --branch ${trustee_version} --depth 1 https://github.com/confidential-containers/trustee.git && cd trustee
```
dcmiddle, line 122: this could be one exception where we want the version specified as written here.

(probably less nice, but we could just say clone the trustee version associated with this release.)

confidential-containers/attestation.rst
dcmiddle: +1

dcmiddle (commented):
Generally lgtm! Good additions on specificity for component versions.

Nit on whether we need to always mention the version in prose.

Unclear on whether we are recommending helm or docker compose. Might be this RA release we are sticking with docker compose, but future release will recommend helm?

manuelh-dev:
> One small request, can you update the release-notes.rst with something like a [Post-Release Documentation Updates](https://github.com/NVIDIA/cloud-native-docs/blob/main/gpu-operator/release-notes.rst?plain=1#L191) section? My concern is to prevent gaslighting a reader who notices changes w/o a record other than commit history.

100%, thanks - I also merged #479 and #480 without adding these. I will add a section which also includes the updates made by these two PRs.

confidential-containers/attestation.rst
manuelh-dev: Thank you, agree'd, I am doing no good here. The goal is version-pinning for our version matrix and for the *actual instructions* (see further below, to assure we use the right commands) but adding these version references everywhere seems redundant. I am entirely removing this added sentence, and I am removing the ${trustee_version} variable from line 71

confidential-containers/attestation.rst
manuelh-dev: @dcmiddle - yes, and we should also adopt this in https://confidentialcontainers.org/docs/examples/nvidia-nim-confidential-gpu-attestation/

The question here is: should we address this as a separate commit in this PR, or make a new PR for this?

@fitzthum - we are generally ready to make the switch here, right? For https://confidentialcontainers.org/docs/examples/nvidia-nim-confidential-gpu-attestation/ I was still waiting for a release which contains: https://github.com/confidential-containers/trustee/pull/1541 ... but I guess on the page here, we don't need to wait?

confidential-containers/attestation.rst
manuelh-dev: My vote: Leave line 123 and 118 as is. I think these instructions should "just work", so the `trustee_version` here and even the version reference on line 118 seem OK to me - here we deal with detailed instructions, not with regular text paragraphs.

confidential-containers/attestation.rst
manuelh-dev: Thanks! This indeed sounds really bad. For now, my best idea is to rephrase this within an additional note:
```
.. note::

The Trustee ${trustee_version} `docker-compose.yml <https://github.com/confidential-containers/trustee/blob/${trustee_version}/docker-compose.yml>`_ still references ``:latest`` images for KBS, AS, and RVPS.
   While cloning that tag pins the compose configuration to a released version, ``docker compose up -d`` still pulls whatever ``:latest`` resolves to at that time
```
@fitzthum - would request your help here. This is related to switching to the helm chart, and/or, to making changes in docker-compose.yml for better pinning.

confidential-containers/attestation.rst
```diff
@@ -184,21 +189,21 @@ Step 3: Install the KBS Client Tool
 
 The KBS client tool, ``kbs-client``, is distributed as a container artifact in the Confidential Containers GitHub Container Registry.
 This tool is mainly used for configuring Trustee.
-
-Pull the ``kbs-client`` artifact into the current directory with ORAS.
+The registry does not publish a ``${trustee_version}`` tag for this artifact.
```
manuelh-dev, line 203: @fitzthum for viz

confidential-containers/confidential-containers-deploy.rst
manuelh-dev: @mikemckiernan @fidencio - this was actually even worse, ptal at my latest commit: https://github.com/NVIDIA/cloud-native-docs/pull/494/changes/089709310cf31cc1fa6e69418fca894bda3b68f3

This should rectify the current state. I am aware of @fidencio 's comment here: https://github.com/NVIDIA/cloud-native-docs/pull/494#discussion_r4017754223

Let's circle back on this, but for now, if we want to keep the current behavior, the new commit should correct things.

dcmiddle:
> One small request, can you update the release-notes.rst with something like a [Post-Release Documentation Updates](https://github.com/NVIDIA/cloud-native-docs/blob/main/gpu-operator/release-notes.rst?plain=1#L191) section? My concern is to prevent gaslighting a reader who notices changes w/o a record other than commit history.

The post-release section was already there. In fact, it was always there.

confidential-containers/attestation.rst
```diff
@@ -142,6 +142,11 @@ Start the Trustee containers in the background.
    On first run, ``docker compose up -d`` pulls the KBS, AS, and RVPS images before starting them.
    This step can take several minutes. The command returns after the containers start. The services may need an additional few seconds to become ready to accept requests.
 
+.. note::
+
+   The Trustee ${trustee_version} `docker-compose.yml <https://github.com/confidential-containers/trustee/blob/${trustee_version}/docker-compose.yml>`_ still references ``:latest`` images for KBS, AS, and RVPS.
+   While cloning that tag pins the compose configuration to a released version, ``docker compose up -d`` still pulls whatever ``:latest`` resolves to at that time.
```
manuelh-dev, line 148: @fitzthum - under discussion, I think for now we want some guidance here on how to change that to the release tag for now

confidential-containers/attestation.rst
fitzthum: Yeah, Helm chart should be ready, especially with the upcoming v0.23.0 release, but idk if we need to make the switch in this PR. I think it's more important to provide some pointers for how the user can patch the image version (with either deployment type).

confidential-containers/attestation.rst
fitzthum, line 148: Let's add some guidance for how to patch the image. We will be able to remove this in the future when we change the Trustee release process.

confidential-containers/attestation.rst
fitzthum: In theory we should be creating a release kbs-client image like we do for all our other images, but looks like we haven't been doing that. I've added to the list of things for the new release process.

fitzthum (commented):
Nice. We can just get the version stuff hammered out with hashes for now. The next CoCo release (v0.24.0) should improve the situation.

confidential-containers/attestation.rst
manuelh-dev: As pointed out in other comments, I will address this. Resolving this comment in favor of the sibling comment

confidential-containers/attestation.rst
manuelh-dev: Noted as an action item. We will make this switch in a new, dedicated PR. Resolving this comment.

confidential-containers/confidential-containers-deploy.rst
manuelh-dev: We discussed to make this change within this PR. I will add a new commit.

repo.toml
manuelh-dev: Per internal discuss, we will leave the versioning as is for now and reconsider when incrementing the version of the reference architecture next time.

confidential-containers/attestation.rst
manuelh-dev: Got guidance here, will provide user instructions on pinning these as this is an important aspect. Resolving, we have a sibling comment

confidential-containers/supported-platforms.rst
manuelh-dev: Per discussion, we will revisit changes with regards to this versioning in a future increment of the RA

confidential-containers/attestation.rst
manuelh-dev: @fitzthum ptal, updated with new concrete pins for the time being - in a future time, we can just remove the `.. note::` on this

confidential-containers/confidential-containers-deploy.rst
manuelh-dev: I am thinking, the scope of this PR increases too much. En contraire to my comment yesterday, I now believe it is better to make this behavioral change in a separate PR.

confidential-containers/attestation.rst
fitzthum: Looks good. We will be able to do this a bit more cleanly when we switch to Helm chart and then before too long we won't need to do it at all.

fitzthum (approved):
LGTM

PR #498, opened 2026-09-17
feat: Updates for 1.20.1

mikemckiernan:


container-toolkit/cdi-support.md
```diff
@@ -251,6 +251,16 @@ Containers can request specific NVIDIA IMEX channels with the
 and validation errors, refer to
 [Requesting IMEX Channels](docker-specialized.md#requesting-imex-channels).
 
+(using-cdi-with-mig-management-devices)=
+
+### MIG Management Devices
+
+JIT-CDI mode injects MIG management capability device nodes when the `NVIDIA_MIG_CONFIG_DEVICES` or `NVIDIA_MIG_MONITOR_DEVICES` environment variable is set to `all` in a privileged container.
+This injects the `/dev/nvidia-caps/` device nodes that tools such as `nvidia-smi mig` need to create, destroy, and monitor MIG partitions.
+The container must have `CAP_SYS_ADMIN`, and the host must use cgroup v2.
```
cdesiniotis: IIRC this change does not depend on cgroup v2, so I would prefer to omit this clause. @henry118 could you confirm?
```suggestion
The container must have `CAP_SYS_ADMIN`.
```

cdesiniotis (commented):
Great work as always @mikemckiernan!

container-toolkit/cdi-support.md
henry118: Strictly speaking, `CAP_SYS_ADMIN` is the only requirement here - not even "privileged" container. 

Note that when a container is "privileged" (either with `--privileged` docker CLI option, or `privileged: true` in k8s pod spec), all devices under `/dev` will be automatically injected by runtime (containerd) by default, regardless the toolkit.

container-toolkit/docker-specialized.md
```diff
@@ -139,6 +139,38 @@ identifies the invalid or missing channel. Inspect
...
+Use the `NVIDIA_MIG_CONFIG_DEVICES` and `NVIDIA_MIG_MONITOR_DEVICES` environment variables to inject MIG management capability device nodes into a container.
+These device nodes are in the `/dev/nvidia-caps/` directory and enable tools such as `nvidia-smi mig` to create, destroy, and monitor MIG partitions from within the container.
+
+Set either variable to `all` to inject the corresponding capability devices.
+No other value is supported.
+
+The following command injects both MIG config and monitor devices:
+
+```console
+$ docker run --rm --runtime=nvidia --privileged \
```
henry118: So here it's better to use `--cap-add=SYS_ADMIN` option instead of `--privileged`. Since mig config/monitor devices will be always available to the container in the latter case.

container-toolkit/release-notes.md
```diff
@@ -8,6 +8,66 @@
...
 This document describes the new features, improvements, fixes and known issues for the NVIDIA Container Toolkit.
 
+## NVIDIA Container Toolkit 1.20.1
+
+This release of the NVIDIA Container Toolkit `v1.20.1` is a bugfix release.
+
+### Fixes and Features
+
+- CDI specifications and JIT-CDI mode now inject MIG management capability devices into containers.
+  Set the `NVIDIA_MIG_CONFIG_DEVICES` or `NVIDIA_MIG_MONITOR_DEVICES` environment variable to `all` in a privileged container to inject the `/dev/nvidia-caps/` device nodes for MIG partition management.
```
henry118: Same here about the "privileged" wording

cdesiniotis (approved):

